PR #4, opened 2023-01-09
function listings in the Find section

topepo:
The code to discover and format relevant functions has been overhauled and should report many more objects: 

 - reverse dependency checks are used to find packages with relevant methods
 - package sources are downloaded and pkgdown is used to parse the documents
 - urlcheck attempts to determine if any package URLs are for pkgdown sites to find the best link

RData files are created in the appropriate directories. These will be updated manually (instead of at render time but we can change that). 

**broom** methods are now included!


EmilHvitfeldt (commented):
I just want to put it down in writing, that we need to find a way to deal with really long variable names 😃 
Screen Shot 2023-01-12 at 12 06 26 AM

Secondly, should be have a way to collapse/hide S3 generics? Like here, is is important to show that `apd_pca()` have all these methods?

Screen Shot 2023-01-12 at 12 06 52 AM

And this is more of a styling/future thing. But it would be cool if we could differentiate between functions and S3 methods

Screen Shot 2023-01-12 at 12 19 24 AM

hfrick (commented):
I second Emil's comments! 

Additionally, where should that link for `model arguments` lead? It did nothing (noticeable) for me when I tried it out.
Screenshot 2023-01-13 at 11 51 41

I generally find the [parsnip reference page](https://parsnip.tidymodels.org//reference/index.html#models) easier to parse. Is the main advantage of this table/search that it also lists packages like modeltime?

topepo:
> Secondly, should be have a way to collapse/hide S3 generics? Like here, is is important to show that apd_pca() have all these methods?

>  But it would be cool if we could differentiate between functions and S3 methods

It's difficult to get information about S3 methods as-is. I don't install the packages, just download the sources and use pkgdown to process them. 

Some S3 methods can't really be figured out unless the packages are installed. For example, the `check_*` functions in the recipes package don't export a lot of their methods so the don't show up in the namespace file. Operating on the string is hard too since the `"."` is overload in naming (e.g. `print.data.frame`). 

We could group them by their Rd file name and this only partially helps since people (incl us) sometimes put methods in a different man page than the generic. 

_If you have the time_, feel free to hack away at it. 

> Additionally, where should that link for model arguments lead? It did nothing (noticeable) for me when I tried it out.

That should be take out.

topepo:
Overall though, please propose edits/`pr_fetch()` if you have good ideas for the `DT` code or anything else.

topepo:
> I just want to put it down in writing, that we need to find a way to deal with really long variable names

Agreed. 

This reminds me... should we show functions beginning with periods?

EmilHvitfeldt:
> This reminds me... should we show functions beginning with periods?

I say no. We talked about it, and those are planned as internal

make_function_lists.R
```diff
@@ -0,0 +1,256 @@
...
+    "workflowsets", "yardstick")
+
+
+tidymodels_functions <- 
+  map_dfr(
+    all_tm, 
+    get_pkg_info, 
+    .progress = TRUE
+  ) %>% 
+  sort_out_urls() %>%
```
simonpcouch, line 175:
```suggestion
  sort_out_urls() %>% 
  filter(grepl("^\\.", functions)) %>%
```

This code should do the trick to remove functions marked as internal👍

simonpcouch (commented):
Jaw actually dropped when poking around the new “Search broom methods” table. This is great.🙂

Agreed that I wish tooling for S3 could be more helpful here, but I think this is fine as-is, especially since the generic seems to most always precede the methods. 

There are already a good few columns in the parsnip table, but I initially expected to see a “package” column—this is often the information that I’m looking for when I’m navigating model type help-files.

The “package” field in the “all of tidymodels” table has a drop-down/scrollable entry list, which is nice, and I think due to that column being a factor rather than character. It’d be nice to have this for model types, engines and modes as well—could we convert those to factor? The remaining string columns make sense as strings.

PR #19, opened 2023-06-13
add code to grey out part of s3 method

EmilHvitfeldt:
I really like this change, it removes some of a eye clutter of this page

Screenshot 2023-06-13 at 11 54 23 AM


find/all/index.qmd
```diff
@@ -17,6 +17,212 @@ tidymodels_prefer()
...
+# Ran in `make_function_lists.R`
+# tidymodels_functions |>
+#   filter(stringr::str_detect(functions, "\\.")) |>
+#   mutate(s3 = stringr::str_remove(functions, "\\..*")) |>
+#   filter(s3 != "") |>
+#   relocate(s3, functions) |>
+#   arrange(s3) |>
+#   View()
+
+s3generics <- c(
```
simonpcouch, line 32: Why is it that we inline this rather than using the results from above? Seems initially like a higher maintenance burden, though I'm totally on board if there's more to this than I'm wrapping my head around.

simonpcouch (approved):
I think this is a really good move! Definitely a helpful cue.

find/all/index.qmd
EmilHvitfeldt: Because there are some false positives for the above results. So I'm manually removing those 🧹 

It will have to be rerun every once in a while to update. But the way the original code is run, there isn't a way to make it foolproof without major overtaking

simonpcouch: Gotcha, that makes sense.

PR #23, opened 2023-06-29
Conformal inference documentation 

topepo:
For regression models.


learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+
+reference_data %>% 
+  ggplot(aes(x = data)) +
+  geom_line(stat = "density") + 
+  geom_vline(xintercept = quants, col = "red", lty = 2)
+```
+
+If we were to get a new sample beyond these data, we would be able to say that we are about 90% sure that the new sample does not conform to the original distribution. This works under the assumption that the data are exchangeable. 
+
+We can apply this relatively simple idea to model predictions. Suppose we have a model created on a numeric outcome. If we make predictions on a data set we can compute the model residuals and create a sort of reference error distribution. If we compute a prediction on a new unknown sample, we could use center this reference distribution around its predicted value. For some confidence level, we now know the range of predicted values that "conform" to the reference distribution. That range can define our prediction interval.
```
EmilHvitfeldt:
```suggestion
We can apply this relatively simple idea to model predictions. Suppose we have a model created on a numeric outcome. If we make predictions on a data set we can compute the model residuals and create a sort of reference error distribution. If we compute a prediction on a new unknown sample, we could center this reference distribution around its predicted value. For some confidence level, we now know the range of predicted values that "conform" to the reference distribution. That range can define our prediction interval. 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+  ggplot(aes(x = data)) +
+  geom_line(stat = "density") + 
+  geom_vline(xintercept = quants, col = "red", lty = 2)
+```
+
+If we were to get a new sample beyond these data, we would be able to say that we are about 90% sure that the new sample does not conform to the original distribution. This works under the assumption that the data are exchangeable. 
+
+We can apply this relatively simple idea to model predictions. Suppose we have a model created on a numeric outcome. If we make predictions on a data set we can compute the model residuals and create a sort of reference error distribution. If we compute a prediction on a new unknown sample, we could use center this reference distribution around its predicted value. For some confidence level, we now know the range of predicted values that "conform" to the reference distribution. That range can define our prediction interval. 
+
+There are a variety of ways to apply this concept (which is unsurprisingly more complex than that the above description). The probably package has implemented a few of them for regression models (classification methods are coming in the future).
```
EmilHvitfeldt:
```suggestion
There are a variety of ways to apply this concept (which is unsurprisingly more complex than that the above description). The probably package has implemented a few of these.
```

The sentence works for a blogpost/announcement. But I think it should be written more generally so we don't have to go back and update this page once we add classification methods

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+      y = y + rnorm(n, sd = std_dev)
+    )
+}
+
+n <- 1000
+set.seed(8383)
+train_data <- make_data(n)
+
+train_data %>% 
+  ggplot(aes(x, y )) +
```
EmilHvitfeldt:
```suggestion
  ggplot(aes(x, y)) + 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+
+To get predictions on new data, we use the standard `predict()` method on this object: 
+
+
+```{r}
+#| label: split-pred
+
+# Produce 90% prediction intervals
+test_split_res <- 
+  predict(split_int, test_data, level = 0.90) %>%
```
EmilHvitfeldt, line 172: This smells like it should have a `augment()` method

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+    data = nnet_pred, aes(y = .pred),
+    linewidth = 1, col = "blue"
+  ) +
+  geom_ribbon(aes(ymin = .pred_lower, ymax = .pred_upper), 
+              col = "#D95F02", linewidth = 3 / 4, fill = NA)
+```
+
+At each point, the interval length is `r round(mean(test_cv_res$.pred_upper - test_cv_res$.pred_lower), 2)`; the previous split conformal interval width was `r round(mean(test_split_res$.pred_upper - test_split_res$.pred_lower), 2)`. This is due to the training set being larger than the calibration set. 
+
+Note that the coverage is a little better too:
```
EmilHvitfeldt: Could you clarify what you mean by "better" here. I read it as "closer to 90%", is that correct?

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+```
+
+Let's create new data sets and re-fit our model: 
+
+```{r}
+#| label: variable-data
+set.seed(7292)
+train_variable_data  <- make_variable_data(n)
+cal_variable_data  <- make_variable_data(250)
+test_variable_data <- make_variable_data(10000)
```
EmilHvitfeldt: nit-pick: either align all the `<-` or none of them 😄

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+```
+
+While the coverage is good (`r round(coverage(test_cv_variable_res)$coverage, 1)`%), there are obvious areas where it is either too wide or too narrow. 
+
+Conformalized quantile regression ([Romano _et al_ (2019)](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Conformalized+quantile+regression%22)) is a method to produce intervals that can properly scale the intervals based on what was observed in the training data. It fits an initial [quantile regression](https://en.wikipedia.org/wiki/Quantile_regression) model to do so and also required a split data set, such as our calibration data. 
+
+The original work used basic quantile regression models. For tidymodels, we'll use [quantile regression forests](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Quantile+Regression+Forests%22). The reason being that some models require nonlinear terms, interactions, and other features to model the data adequately. Users would have to have a parallel modeling process for that model. Random forests for quantile regression are very adaptable, are low maintenance, and are robust to tuning parameter specification. There is a minor downside, as seen below.
```
EmilHvitfeldt: I don't know if this is a good change, but it sounds better to me

```suggestion
The original work used basic quantile regression models. For tidymodels, we'll use [quantile regression forests](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Quantile+Regression+Forests%22). The reason being that some models require nonlinear terms, interactions, and other features to model the data adequately. Users would have to have a parallel modeling process for that model. Random forests for quantile regression are very adaptable, low maintenance, and are robust to tuning parameter specification. There is a minor downside, as seen below. 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+  geom_point(aes(y = y), alpha = 1 / 20) +
+  geom_line(
+    data = nnet_variable_pred, aes(y = .pred),
+    linewidth = 1, col = "blue"
+  ) +
+  geom_ribbon(aes(ymin = .pred_lower, ymax = .pred_upper), 
+              col = "#D95F02", linewidth = 1, fill = NA)
+```
+
+While the coverage is good (`r round(coverage(test_cv_variable_res)$coverage, 1)`%), there are obvious areas where it is either too wide or too narrow.
```
EmilHvitfeldt: Is there a reason why you stopped running `coverage(test_cv_variable_res)` for people to see?

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,441 @@
...
+
+## Coverage
+
+Speaking of coverage, there's a [GitHub repo](https://github.com/topepo/conformal_sim) where we did simulations for these methods to determine their average coverage. The repository's readme file has a summary. 
+
+## Looking Forward
+
+This is a growing research area; we will do our best to implement important new methods as they are created. 
+
+Also, there are a multitude of methods for classification models. Those are next on our list.
```
EmilHvitfeldt: Same as before, I don't think this line to good to have on a long term basis

EmilHvitfeldt (approved):
Looking good! a number of comments

EmilHvitfeldt:
Once we are happy with it, I can re-render it with the right quarto version so the diff is smaller

learn/models/conformal-regression/index.qmd
topepo: We don't have one implemented yet but that's a great idea.

learn/models/conformal-regression/index.qmd
topepo: Just to be more concise.

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+registerDoMC(cores = parallel::detectCores())
+
+# ------------------------------------------------------------------------------
+
+pkgs <- c("tidymodels", "probably", "nnet")
+```
+
+`r article_req_pkgs(pkgs)` The probably package should be version 1.0.1 or greater.
+
+What is [conformal inference](https://en.wikipedia.org/wiki/Conformal_prediction)? It is a collection of statistical methods that are mostly used to construct prediction intervals (or prediction sets) for any type of regression or classification model. The basic idea revolves around some theory on how to construct probability statements for a population of data points.
```
hfrick: That last sentence about the basic idea is very broad/vague

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+#| fig-height: 4
+quants <- quantile(reference_data$data, probs = c(0.05, 0.95))
+
+reference_data %>% 
+  ggplot(aes(x = data)) +
+  geom_line(stat = "density") + 
+  geom_vline(xintercept = quants, col = "red", lty = 2)
+```
+
+If we were to get a new sample beyond these data, we would be able to say that we are about 90% sure that the new sample does not conform to the original distribution. This works under the assumption that the data are exchangeable.
```
hfrick:
```suggestion
If we were to get a new sample beyond these boundaries, we would be able to say that we are about 90% sure that the new sample does not conform to the original distribution. This works under the assumption that the data are exchangeable. 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+  ggplot(aes(x = data)) +
+  geom_line(stat = "density") + 
+  geom_vline(xintercept = quants, col = "red", lty = 2)
+```
+
+If we were to get a new sample beyond these data, we would be able to say that we are about 90% sure that the new sample does not conform to the original distribution. This works under the assumption that the data are exchangeable. 
+
+We can apply this relatively simple idea to model predictions. Suppose we have a model created on a numeric outcome. If we make predictions on a data set we can compute the model residuals and create a sort of reference error distribution. If we compute a prediction on a new unknown sample, we could center this reference distribution around its predicted value. For some confidence level, we now know the range of predicted values that "conform" to the reference distribution. That range can define our prediction interval. 
+
+There are a variety of ways to apply this concept (which is unsurprisingly more complex than that the above description). The probably package has implemented a few of these.
```
hfrick:
```suggestion
There are a variety of ways to apply this concept (which is unsurprisingly more complex than the above description). The probably package has implemented a few of these.
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+            linewidth = 1, col = "blue")
+```
+
+Let's examine three methods^[A fourth method is implemented for "full conformal prediction." It was developed mainly as a proof of concept. While it is effective, it is _very_ computationally expensive.] to produce prediction intervals: 
+
+## Split Conformal Inference
+
+The most straightforward approach is reserving some data for estimating the residual distribution. We know that simply re-predicting the training set is **a bad idea**; the residuals would be smaller than they should be since the same data are used to create and evaluate the model. 
+
+Let's simulate another data set containing 250 samples and call that the "calibration set". These data can be predicted and the corresponding residuals can be used to define what conforms to model. We'll also create a large test set to see if we've done a good job.
```
hfrick:
```suggestion
Let's simulate another data set containing 250 samples and call that the "calibration set". These data can be predicted and the corresponding residuals can be used to define what conforms to the model. We'll also create a large test set to see if we've done a good job. 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+  signif(3)
+```
+
+Since we know the outcome values, we can compute the coverage for this particular data set. Since we created a 90% prediction interval, about 90% of the outcomes should be within our bounds. Let's create a function and apply it to these data:
+
+```{r}
+#| label: split-coverage
+coverage <- function(x) {
+  x %>% 
+    mutate(in_bound = .pred_lower < y & .pred_upper > y) %>%
```
hfrick: Should this be inclusive of the boundaries?

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+  geom_point(aes(y = y), alpha = 1 / 10) +
+  geom_line(data = nnet_pred, aes(y = .pred),
+            linewidth = 1, col = "blue")
+```
+
+Let's examine three methods^[A fourth method is implemented for "full conformal prediction." It was developed mainly as a proof of concept. While it is effective, it is _very_ computationally expensive.] to produce prediction intervals: 
+
+## Split Conformal Inference
+
+The most straightforward approach is reserving some data for estimating the residual distribution. We know that simply re-predicting the training set is **a bad idea**; the residuals would be smaller than they should be since the same data are used to create and evaluate the model.
```
hfrick, line 143: It'd be nice if this section makes it more clear why this method is called split conformal inference. I'm guessing it's because we split the data here, where you call it reserving data?

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+  x %>% 
+    mutate(in_bound = .pred_lower < y & .pred_upper > y) %>% 
+    summarise(coverage = mean(in_bound) * 100)
+}
+coverage(test_split_res)
+```
+
+## Using Resampling Results
+
+In a way, the calibration set serves a similar role as a traditional validation set. Since we can just re-predict our training set, we need to evaluate our model on a separate collection of labeled data.
```
hfrick: non-native speaker question: do you serve a role or is that a mash-up of serving a purpose and having a role?

hfrick:
```suggestion
In a way, the calibration set serves a similar role as a traditional validation set. Since we can't just re-predict our training set, we need to evaluate our model on a separate collection of labeled data. 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+    summarise(coverage = mean(in_bound) * 100)
+}
+coverage(test_split_res)
+```
+
+## Using Resampling Results
+
+In a way, the calibration set serves a similar role as a traditional validation set. Since we can just re-predict our training set, we need to evaluate our model on a separate collection of labeled data. 
+
+However, we also know that resampling methods can serve the same purpose. During resampling, we can compute a set of predictions that were not used to fit the model. For example, in 10-fold cross-validation, there are 10 collections of held-out predictions that do not suffer from the severe bias that occurs when we simple re-predict the training set.
```
hfrick, line 219: The "however" feels a bit out of place to me

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+folds <- vfold_cv(train_data)
+
+nnet_rs <- 
+  nnet_wflow %>% 
+  fit_resamples(folds, control = ctrl)
+
+collect_metrics(nnet_rs)
+```
+
+The model has an estimated root mean squared error of `r collect_metrics(nnet_rs) %>% dplyr::filter(.metric == "rmse") %>% pluck("mean") %>% signif(3)` and an estimated R<sup>2</sup> of `r collect_metrics(nnet_rs) %>% dplyr::filter(.metric == "rsq") %>% pluck("mean") %>% signif(3)`.
```
hfrick, line 245: Is this to get a general sense for how the model is doing? If so, it might be helpful to say that, so that people don't assume that this is necessary for getting prediction intervals.

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+The original work used basic quantile regression models. For tidymodels, we'll use [quantile regression forests](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Quantile+Regression+Forests%22). The reason being that some models require nonlinear terms, interactions, and other features to model the data adequately. Users would have to have a parallel modeling process for that model. Random forests for quantile regression are very adaptable, low maintenance, and are robust to tuning parameter specification. There is a minor downside, as seen below. 
+
+The function for this, `int_conformal_quantile()`, has a slightly different interface than the previous functions: 
+
+* It requires both the training and calibration sets.
+* The confidence level must be set in advance. 
+
+We'll also pass an argument to the `quantregForest()` function that will use 2,000 trees to make the forest:
```
hfrick:
```suggestion
We'll also pass an argument to the `quantregForest()` function that we'll use 2,000 trees to make the forest: 
```

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+    data = nnet_variable_pred, aes(y = .pred),
+    linewidth = 1, col = "blue"
+  ) +
+  geom_ribbon(aes(ymin = .pred_lower, ymax = .pred_upper), 
+              col = "#D95F02", linewidth = 3 / 4, fill = NA)
+```
+
+The bumpiness of the bound is the consequence of using a tree-based model for the quantiles. Despite that, the intervals do adapt their widths to suit the data. The coverage is also close to the target: `r round(coverage(test_quant_res)$coverage, 1)`%. 
+
+## Coverage
```
hfrick, line 478: This sounds a lot more casual than other parts of the text.

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+```
+
+While the coverage is good (`r round(coverage(test_cv_variable_res)$coverage, 1)`%), there are obvious areas where it is either too wide or too narrow. 
+
+Conformalized quantile regression ([Romano _et al_ (2019)](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Conformalized+quantile+regression%22)) is a method to produce intervals that can properly scale the intervals based on what was observed in the training data. It fits an initial [quantile regression](https://en.wikipedia.org/wiki/Quantile_regression) model to do so and also required a split data set, such as our calibration data. 
+
+The original work used basic quantile regression models. For tidymodels, we'll use [quantile regression forests](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=%22Quantile+Regression+Forests%22). The reason being that some models require nonlinear terms, interactions, and other features to model the data adequately. Users would have to have a parallel modeling process for that model. Random forests for quantile regression are very adaptable, low maintenance, and are robust to tuning parameter specification. There is a minor downside, as seen below.
```
hfrick: I really like this. It might be worth it to be a little bit more explicit about that that minor downside is, i.e. refer back to this sentence when you talk about the bumpiness of the intervals?

learn/models/conformal-regression/index.qmd
hfrick: I'm getting stuck on this sentence.

If we are talking confidence intervals, I think that sentence is skirting between "the interval covers the parameter" and "the parameter lies in the interval". But since we are talking about prediction intervals, does this still apply? 

What I'm trying to say: is this sentence statistically correct? I can't judge that immediately and my training on how to talk about p-values is kicking in - but it might be misplaced here.

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+The bumpiness of the bound is the consequence of using a tree-based model for the quantiles. Despite that, the intervals do adapt their widths to suit the data. The coverage is also close to the target: `r round(coverage(test_quant_res)$coverage, 1)`%. 
+
+## Coverage
+
+Speaking of coverage, there's a [GitHub repo](https://github.com/topepo/conformal_sim) where we did simulations for these methods to determine their average coverage. The repository's readme file has a summary. 
+
+## Looking Forward
+
+This is a growing research area; we will do our best to implement important new methods as they are created.
```
hfrick: I think we usually don't put such forward-looking statements in the tidymodels.org articles?  That seems much more suited for a blog post, maybe with a link to the GH issue to suggest particularly great new methods.

learn/models/conformal-regression/index.qmd
```diff
@@ -0,0 +1,437 @@
...
+
+## Coverage
+
+Speaking of coverage, there's a [GitHub repo](https://github.com/topepo/conformal_sim) where we did simulations for these methods to determine their average coverage. The repository's readme file has a summary. 
+
+## Looking Forward
+
+This is a growing research area; we will do our best to implement important new methods as they are created. 
+
+If you are interested and would like to learn more, try these resources:
```
hfrick, line 484: While it's great to have places to learn more, this reminds me of "Never be the keeper of the list".

hfrick (approved):
One very stats-y question and a couple of comments. I like the visualizations, I think that makes it all pretty easy to follow!

learn/models/conformal-regression/index.qmd
topepo: I rewrote it

learn/models/conformal-regression/index.qmd
topepo: I rewrote this (emphasis added here to show the differences): 

> If we compute a prediction on a new unknown sample, we could center this reference distribution around its predicted value. For some _significance_ level, we now know the range of _sample_ values that "conform" to the _variance seen in the_ reference distribution. That range can define our prediction interval.

topepo: The use of "predicted values" instead of "sample values" was incorrect.

PR #25, opened 2023-06-30
Overview and details articles on survival metrics

hfrick:
Here are the two articles on survival metric, moved over from 
- https://github.com/tidymodels/tidymodels.org-legacy/pull/283 and
- https://github.com/tidymodels/tidymodels.org-legacy/pull/282

@EmilHvitfeldt could you glance at this PR with an eye towards "looks alright wrt to quarto"? 

I used quarto v1.3.433 which is the latest release version.


EmilHvitfeldt:
If possible, I would like for a {parsnip} CRAN release before this post is added. As for right now, it would be the only article that requires a dev core package. Ideally a CRAN release of {censored} as well.

topepo:
We want the survival work in tune (and other packages) to sit in devel for a while so that external folks can play with it.

PR #26, opened 2023-06-30
Add workflow section to README

hfrick:
Here's the workflow section we talked about!

When I render it, I get a README.html (and `README_files/` folder) which were not previously part of the repo. Should they be checked in or were they left out on purpose?


EmilHvitfeldt:
The readme is a .md file so it doesn't need to be rendered :)

hfrick:
Right! 🤦‍♀️ 😆  Thanks!

PR #50, opened 2023-11-20
Fix typo in neural network example

michaelgrund:
The code example defines 10 hidden units while the text only contains a number of 5.


EmilHvitfeldt:
thank you @michaelgrund !!!

PR #51, opened 2023-11-21
Fix typo in regression models example

michaelgrund:


EmilHvitfeldt:
thank you @michaelgrund !!!

PR #48, opened 2023-11-08
add applied fairness analysis with yardstick

simonpcouch:
An example of an applied fairness analysis with yardstick based on the GPT detector data in the detectors R package.

This is 1 of 2 applied fairness analyses I'd like to add to tidymodels.org:

* This analysis uses 1 pre-defined metric, only uses yardstick, and focuses on socially held ideas of fairness and their relationship with the metrics.
* The analysis in a later PR uses several metrics and resamples them with tune.

**This is intended to be merged only _after_ the release of yardstick 1.3.0 on CRAN.** Rendered with the needed dev versions, for now.

Edit: [here's where the new article is situated](https://deploy-preview-48--tidymodels-org.netlify.app/learn/work/fairness-detectors/) in the deploy preview.

PR #52, opened 2023-11-29
add applied fairness analysis with tune

simonpcouch:
An example of an applied fairness analysis with tune and yardstick based on hospital readmission data.

This is 2 of 2 applied fairness analyses I'd like to add to tidymodels.org, the first in #48.

**This is intended to be merged only _after_ the release of yardstick 1.3.0 on CRAN.** Rendered with the needed dev versions, for now.

[Here's where the new article is situated](https://deploy-preview-52--tidymodels-org.netlify.app/learn/work/fairness-readmission/) in the deploy preview.


installs.R
```diff
@@ -47,6 +47,7 @@ packages <- c(
   "rules",
   "scales",
   "sessioninfo",
+  "simonpcouch/readmission",
```
simonpcouch: Headed to CRAN once we feel this article is in a good spot.

learn/work/fairness-readmission/index.md
```diff
@@ -0,0 +1,739 @@
...
+#>  purrr       * 1.0.2      2023-08-10 [1] CRAN (R 4.3.0)
+#>  readmission * 0.0.3      2023-11-29 [1] Github (simonpcouch/readmission@7d1a793)
+#>  recipes     * 1.0.8      2023-08-25 [1] CRAN (R 4.3.0)
+#>  rlang         1.1.2      2023-11-04 [1] CRAN (R 4.3.1)
+#>  rsample     * 1.2.0      2023-08-23 [1] CRAN (R 4.3.0)
+#>  tibble      * 3.2.1      2023-03-20 [1] CRAN (R 4.3.0)
+#>  tidymodels  * 1.1.1      2023-08-24 [1] CRAN (R 4.3.0)
+#>  tune        * 1.1.2.9000 2023-11-27 [1] Github (tidymodels/tune@3509577)
+#>  workflows   * 1.1.3      2023-02-22 [1] CRAN (R 4.3.0)
+#>  yardstick   * 1.2.0.9001 2023-11-27 [1] Github (tidymodels/yardstick@2f556df)
```
simonpcouch: To be updated when dev tune + yardstick on CRAN.

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+
+To choose a model that performs well both with respect to a typical performance metric like `roc_auc()` and the fairness metrics we've chosen, we will:
+
+1) Find the "best" model with respect to `roc_auc()`.
+2) Choose the most fair model with respect to `equalized_odds()` among the models within one standard error of `roc_auc()` to the model we identified in step 1.
+
+We can do so by first aggregating performance metrics from the `"age_bt"` result with `collect_metrics()` and then using some tidyr and dplyr for data manipulation:
+
+<!-- 
+TODO: it'd be great if `select_by_one_std_err()` could "just work" here, i.e.
```
simonpcouch: https://github.com/tidymodels/tune/issues/771

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+
+This article will demonstrate a fairness-oriented workflow for training a machine learning model to identify high-risk patients. Throughout the model development process, we'll consider the social impacts that different modeling decisions may have when such a model is deployed in context.
+
+## Setup
+
+The data we'll use in this analysis is a publicly available database containing information on `r format(nrow(readmission), big.mark = ",")` hospital stays from diabetes patients. The data comes from a study by @strack2014, where the authors model the effectiveness of a particular lab test in predicting readmission. A version of that data is available in the [readmission R package](https://simonpcouch.github.io/readmission/):
+
+```{r}
+#| label: "load-readmission"
+#| eval: false
```
topepo: Do you mean to show the data (with the last line)? If not, I'd remove it.

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+```
+
+The first variable in this data, `readmitted`, gives whether the patient was readmitted within 30 days of discharge. We'll use this variable as a proxy for "unmet need for additional care," in that readmission within one month indicates that the patient may have benefited from additional attention during their hospital stay; if a machine learning model consistently identifies lesser need (via prediction of non-readmission) in one subgroup than another, the subgroup allocated lesser need will go without care they'd benefit from. We'd like to train a model that is both fair with regard to how it treats `race` groups and is as performant as possible. The tidymodels framework provides the tools needed to identify such disparities.
+
+Loading needed packages:
+
+```{r}
+#| label: "load-tidymodels"
+library(tidymodels)
+library(baguette)
```
topepo, line 81: Maybe add `tidymodels_prefer()`?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+We'll refer to the `race` groups in this data by their actual value (e.g. `"Caucasian"` rather than Caucasian) so as to not to take for granted the choices that the dataset authors made in choosing these categorizations. Racial categorizations are not stable across time and place---as you read on, consider how a change in the categories used in data collection might affect this analysis [@omi1994].
+:::
+
+A vast majority of patients are labeled `"Caucasian"` (`r round(mean(readmission$race == "Caucasian"), 3)*100`%) or `"African American"` (`r round(mean(readmission$race == "African American"), 3)*100`%). The counts for the remaining racial categorizations are quite a bit smaller, and when we split the data up into resamples, those counts will reduce even further. As a result, the variability associated with the estimates `"Asian"`, `"Hispanic"`, `"Other"`, and `"Unknown"` will be larger than those for `"African American"` and `"Causasian"`. As an example:
+
+```{r}
+#| label: "prop-by-race"
+#| message: false
+#| warning: false
+readmission %>%
```
topepo, line 120: I think that standard errors here are not great. They are complete driven by the magnitude of the proportions. Instead, I'd use a combination of `prop.test()` and `tidy()` to get the confidence intervals back.

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+
+Recoding that data column:
+
+```{r}
+#| label: "collapse-race"
+readmission_collapsed <-
+  readmission %>%
+  mutate(
+    race = case_when(
+      !race %in% c("Caucasian", "African American") ~ "Other",
```
topepo: omg I know it's not needed but the `!race` part is needling me.

```suggestion
      !(race %in% c("Caucasian", "African American")) ~ "Other",
```

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+  step_unknown(all_nominal_predictors()) %>%
+  step_YeoJohnson(all_numeric_predictors()) %>%
+  step_normalize(all_numeric_predictors()) %>%
+  step_dummy(all_nominal_predictors())
+```
+
+The other preprocessor that we'll try encodes the age as a numeric variable rather than the bins as in the source data:
+
+```{r}
+#| label: "recipe-age"
```
topepo, line 251: I would add some comments here; it would be intimidating for folks that are not R experts.

topepo: Also, I would add a better name for the x-axis

topepo: If you are trying to show the relationship between the metrics (i.e. as ROC increses, X decreases), I would use a scatterplot matrix instead.

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+
+We can do so by first aggregating performance metrics from the `"age_bt"` result with `collect_metrics()` and then using some tidyr and dplyr for data manipulation:
+
+<!-- 
+TODO: it'd be great if `select_by_one_std_err()` could "just work" here, i.e.  
+select_by_one_std_err(best_workflow, metric = "roc_auc", desc(equalized_odds))
+-->
+
+```{r}
+#| label: "find-best-params"
```
topepo, line 423: I'm not sure that you want to introduce another concept here but this is a perfect use case for desirability functions.

topepo (commented):
Looks great overall!

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+recipe_age <-
+  recipe(readmitted ~ ., data = readmission) %>%
+  step_mutate(
+    age_num = 
+      age %>% 
+      as.character() %>% 
+      substr(2, 2) %>% 
+      as.numeric() %>%
+      `*`(10) %>% `+`(5)
+  ) %>%
```
EmilHvitfeldt: The age thing right here isn't is a little much for a pipe in my opinion. I would pull it out as a helper function

```suggestion
# specify the age as the midpoint of the range
# rather than as an interval, like in the source data
# e.g. "[10-20]" -> 1 * 10 + 5 -> 15
get_age <- function(x) {
  res <- substr(x, 2, 2)
  res <- as.numeric(res)
  res * 10 + 5
}

recipe_age <-
  recipe(readmitted ~ ., data = readmission) %>%
  step_mutate(
    age_num = get_age(age)
  ) %>%
```

EmilHvitfeldt: OR you could be overly fancy and do

```r
step_mutate(
  age_num = as.numeric(age) * 10 - 5
)
```

learn/work/fairness-readmission/index.qmd
juliasilge: Or should this be `eval: true` to show it?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+
+Let's start off our analysis with some explanatory summarization and plotting. First, taking a look at our outcome variable:
+
+```{r}
+#| label: "count-by-readmitted"
+readmission %>%
+  count(readmitted)
+```
+
+`r round(mean(readmission$readmitted == "Yes"), 3)*100`% of patients readmitted within 30 days after being discharged from the hospital. This is an example of a modeling problem with a _class imbalance_, where one value of the outcome variable is much more common than another. Now, taking a look at the counts of those in each protected class:
```
juliasilge:
```suggestion
`r round(mean(readmission$readmitted == "Yes"), 3)*100`% of patients were readmitted within 30 days after being discharged from the hospital. This is an example of a modeling problem with a _class imbalance_, where one value of the outcome variable is much more common than another. Now, taking a look at the counts of those in each protected class:
```

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+With a sense for the distributions of variables in this dataset, we can move on to splitting data up for modeling.
+
+## Resampling Data
+
+First, splitting data into training and testing:
+
+```{r}
+#| label: "split"
+set.seed(1)
+readmission_splits <- initial_split(readmission_collapsed)
```
juliasilge: Want to consider using `strata` since there is some class imbalance?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+readmission_train <-  training(readmission_splits)
+readmission_test <-   testing(readmission_splits)
+```
+
+We'll set the `r format(nrow(readmission_test), big.mark = ",")`-row `readmission_test` test set to the side for the remainder of the analysis until we compute a final estimate of our performance on the chosen model. Splitting the `r format(nrow(readmission_train), big.mark = ",")` rows of the training data into 10 resamples:
+
+```{r}
+#| label: "resample"
+readmission_folds <- vfold_cv(readmission_train)
```
juliasilge: Use `strata`?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+  )
+
+wflow_set
+```
+
+Each workflow in the workflow set is now ready to be evaluated. We now need to decide how to best evaluate these modeling workflows, though.
+
+### Metrics
+
+The metrics with which we choose to evaluate our models with are the core of our fairness analysis. In addition to the default metrics for classification in tune, `accuracy()` and `roc_auc()`, we'll compute a set of fairness metrics: `equal_opportunity()`, `equalized_odds()`, and `demographic_parity()`.
```
juliasilge:
```suggestion
The metrics with which we choose to evaluate our models are the core of our fairness analysis. In addition to the default metrics for classification in tune, `accuracy()` and `roc_auc()`, we'll compute a set of fairness metrics: `equal_opportunity()`, `equalized_odds()`, and `demographic_parity()`.
```

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+    equalized_odds(race),
+    demographic_parity(race)
+  )
+
+m_set
+```
+
+
+::: callout-note
+The first two inputs, `accuracy()` and `roc_auc()`, are standard [yardstick](https://yardstick.tidymodels.org) metrics. The latter three are also yardstick metrics like any other, though are created using the [_metric factories_](https://yardstick.tidymodels.org/dev/reference/new_groupwise_metric.html) `equal_opportunity()`, `equalized_odds()`, and `demographic_parity()`. When passed a data-column, metric factories output yardstick metrics.
```
juliasilge: This is really nice 👍

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+```
+
+## Model Selection
+
+Now that we've evaluated a number of models with a variety of metrics, we can explore the results to determine our optimal model. Beginning by a quick exploratory plot of our metrics:
+
+<!-- TODO: i'd really like to use `autoplot(wflow_set_fit, rank_metric = "roc_auc")` here, but the plot isn't usable without differentiating between the recipes. -->
+
+```{r}
+#| label: "plot-wflow-set-fit"
```
juliasilge: I do think this plot could be clearer, like @topepo said earlier. The code does look really overwhelming, and the x-axis is kind of awkward. I do wonder about something more like a pairs plot for all the metrics.

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+  ) %>%
+  ggplot(aes(x = .config, y = mean, color = wflow_id)) +
+  facet_wrap(vars(.metric), scales = "free") +
+  geom_point() +
+  scale_x_discrete(breaks = NULL)
+```
+
+Each dot in the plots above represents one configuration of a modeling workflow. For each modeling workflow besides logistic regressions, we've tried 10 unique combinations of hyperparameters. For the workflows using logistic regressions, we've tried only one, since we didn't mark any parameters for tuning using the `tune()` placeholder. The dots are "ranked" by `roc_auc()` in that, in each pane, the left-most models took the highest `roc_auc()` values.
+
+The fairness metrics `demographic_parity()`, `equal_opportunity()`, and `equalized_odds()` all take values very close to zero for many models. Notably, the left-most models (i.e. the most performant models with respect to `roc_auc()`) happen to be among the most fair according to the metrics we've chosen.
```
juliasilge: Can we say more explicitly or earlier that close to zero is more fair?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+#| label: "rank-results"
+rank_results(wflow_set_fit, rank_metric = "roc_auc") %>%
+  filter(.metric == "roc_auc")
+```
+
+As in the plot above, we see that almost all of the most performant model configurations arise from the boosted tree modeling workflow. Let's examine the results for specifically the modeling workflow that encodes age as a number more thoroughly:
+
+```{r}
+#| label: "autoplot-age-bt"
+autoplot(wflow_set_fit, id = "age_bt")
```
juliasilge, line 409: Can you set fig width and height to make the metric labels more readable?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,517 @@
...
+```
+
+The learning rate `learn_rate` seems to have a more pronounced effect on the results metrics than the number of randomly selected predictors `mtry`. As before, we see that the most performant models with respect to `roc_auc()` tend to be the most fair with respect to our fairness metrics. Further, the values of each of the fairness metrics plotted above seem highly correlated. 
+
+From the perspective of a practitioner hoping to satisfy various stakeholders, the fact that these metrics are highly correlated makes the model selection process much easier. We can choose one fairness metric that we'd like to optimize for, and likely end up with a near-optimal configuration for the other metrics as a byproduct.
+
+:::callout-note
+In machine learning fairness, "impossibility thereoms" show that fairness definitions "are not mathematically or morally compatible in general" [@mitchell2021]. More concretely, unless we live in a world with no inequality, there is no way to satisfy many definitions of fairness at once. However, recent research emphasizes that near-fairness among more limited sets of metrics, like the three we've used here, is both possible and relevant [@bell2023].
```
juliasilge: This and the previous paragraph are really nice 👍

juliasilge (commented):
This is, once again, looking fantastic! 🙌 

I left some individual comments, but my one last thought is that I think folks will come away with questions about the accuracy being basically the same as just saying "never readmit". Maybe link off to a different article on class imbalance? Or talk about adding a different metric that is more about the minority class?

learn/work/fairness-readmission/index.qmd
simonpcouch: What are the comments about a plot referring to? 

[EDIT: going to run with these referring to `"plot-wflow-set-fit"`!]

learn/work/fairness-readmission/index.qmd
simonpcouch: Ah, yes! [115a87b](https://github.com/tidymodels/tidymodels.org/pull/52/commits/115a87bc8ac9c045789fa29970a5091889296d16).

learn/work/fairness-readmission/index.qmd
simonpcouch: [115a87b](https://github.com/tidymodels/tidymodels.org/pull/52/commits/115a87bc8ac9c045789fa29970a5091889296d16)!

learn/work/fairness-readmission/index.qmd
simonpcouch: Heard! [115a87b](https://github.com/tidymodels/tidymodels.org/pull/52/commits/115a87bc8ac9c045789fa29970a5091889296d16).

learn/work/fairness-readmission/index.qmd
simonpcouch: Thank you! [115a87b](https://github.com/tidymodels/tidymodels.org/pull/52/commits/115a87bc8ac9c045789fa29970a5091889296d16).

learn/work/fairness-readmission/index.qmd
simonpcouch: Yes! Does a3dfdea do the trick?

learn/work/fairness-readmission/index.qmd
simonpcouch: Yes! 5dce621. Let me know if that does the trick---the metric names are now fully visible but I'm curious how the size looks on others' setups.

learn/work/fairness-readmission/index.qmd
simonpcouch: Thank you! [115a87b](https://github.com/tidymodels/tidymodels.org/pull/52/commits/115a87bc8ac9c045789fa29970a5091889296d16). :)

learn/work/fairness-readmission/index.qmd
simonpcouch: Heard. I guess the big pictures here are

1) The metrics are very correlated, and
2) Some model workflows do better on average than others.

Those could be two separate plots.🏄

simonpcouch: Actually, you're spot on. Let's just plot the metrics pairwise and then print out the most performant models from `roc_auc()` in a table. :) 474ea32!

learn/work/fairness-readmission/index.qmd
simonpcouch: Addressed in [115a87b](https://github.com/tidymodels/tidymodels.org/pull/52/commits/115a87bc8ac9c045789fa29970a5091889296d16). :)

learn/work/fairness-readmission/index.qmd
simonpcouch: Unresolving, as there's an issue with the x axis labels.

simonpcouch: Okay, better. :) 91ca48a.

learn/work/fairness-readmission/index.qmd
simonpcouch: Wrapped my head around the sizing a bit better--91ca48a.

learn/work/fairness-readmission/index.qmd
simonpcouch: Yes, good thought.

In [80fe7e5](https://github.com/tidymodels/tidymodels.org/pull/52/commits/80fe7e5a195413d647ea411b822a2083b39706a2) I added the argument, but it won't be reflected in the `wflow_set_fit` results until I refit the workflow set once the article is otherwise ready to go. :)

learn/work/fairness-readmission/index.qmd
simonpcouch: [80fe7e5](https://github.com/tidymodels/tidymodels.org/pull/52/commits/80fe7e5a195413d647ea411b822a2083b39706a2)!

learn/work/fairness-readmission/index.qmd
simonpcouch: Absolutely. f6842b0. :)

learn/work/fairness-readmission/index.qmd
simonpcouch: I _think_ we're in agreement here that the demonstrated thing is a bad idea. I'm showing that, though the proportions of readmission are quite similar across `race` values, variation in the number of values being used to estimate those proportions drives greater variation of estimates across resamples. The fairness metrics are vulnerable to that variation since they measure differences in proportion in each resample. The `prop.test() %>% tidy()` approach would be one way to address that variation more sensibly, which I've written we ought to do if we decided to retain all of the factor levels, though we don't ultimately do that. 

Do let me know if I'm missing your point. :)

simonpcouch:
re: your general comment, @juliasilge, let me know if [80fe7e5](https://github.com/tidymodels/tidymodels.org/pull/52/commits/80fe7e5a195413d647ea411b822a2083b39706a2) does the trick. Glad to add some discussion about a metric that speaks specifically to performance in the minority class.

juliasilge (approved):
This is looking great! The pairs plot especially is a big improvement.

The only last question/comment I have is whether to in fact add at least a sentence about a metric for the minority class, probably when you are discussing the results. I think it's up for debate whether that's super helpful, though; maybe instead you could link once more to the class imbalance article.

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,516 @@
...
+collect_metrics(final_model)
+```
+
+```{r}
+#| label: "calc-final-metrics"
+#| include: false
+final_metrics <- collect_metrics(final_model)
+```
+
+The model we've selected has near-fairness with respect to the set of metrics we've chosen here. The accuracy of the model is `r round(final_metrics$.estimate[final_metrics$.metric == "accuracy"], 4)*100`%, quite similar to the accuracy that would result if the model just always predicted a patient would not readmit (`r (1 - round(mean(readmission_collapsed$readmitted == "Yes"), 4))*100`%). The `roc_auc()` value `r round(final_metrics$.estimate[final_metrics$.metric == "roc_auc"], 3)` indicates that the model indeed correctly predicts readmission in some cases, though still has a lot of room for improvement.
```
juliasilge: This is the paragraph where I think that _maybe_ there should be one more sentence either linking again to the class imbalance article or talking about a metric choice that could have given a clearer answer on minority class performance. However, use your judgment on whether it does/doesn't help that much.

simonpcouch: You got it! [5293c2d](https://github.com/tidymodels/tidymodels.org/pull/52/commits/5293c2d4f3d3c1d4f486e8a8560e77a96fbb56a4).

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,516 @@
...
+`r round(mean(readmission$readmitted == "Yes"), 3)*100`% of patients were readmitted within 30 days after being discharged from the hospital. This is an example of a modeling problem with a _class imbalance_, where one value of the outcome variable is much more common than another. Now, taking a look at the counts of those in each protected class:
+
+```{r}
+#| label: "count-by-race"
+readmission %>%
+  count(race)
+```
+
+::: callout-note
+We'll refer to the `race` groups in this data by their actual value (e.g. `"Caucasian"` rather than Caucasian) so as to not to take for granted the choices that the dataset authors made in choosing these categorizations. Racial categorizations are not stable across time and place---as you read on, consider how a change in the categories used in data collection might affect this analysis [@omi1994].
```
EmilHvitfeldt: The series of words `so as to not to` is not my favorite

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,516 @@
...
+Most patients in this data are in their 60s and 70s. Emergencies account for most admissions in this data, though many others are from referrals or other sources.
+
+```{r}
+#| label: "plot-insurer"
+readmission_collapsed %>%
+  ggplot(aes(x = insurer)) +
+  geom_bar()
+```
+
+While payment information on most patients is missing, most patients in this data are covered under medicare.
```
EmilHvitfeldt:
```suggestion
While payment information on most patients is missing, most patients in this data are covered under Medicare.
```

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,516 @@
...
+A fitted workflow set looks just like the unfitted workflow set we saw previously, except that information on the tuning process is now stored in the `option` and `result` variables for each modeling workflow:
+
+```{r}
+#| label: "print-wflow-set"
+wflow_set_fit
+```
+
+## Model Selection
+
+Now that we've evaluated a number of models with a variety of metrics, we can explore the results to determine our optimal model. Beginning by a quick exploratory plot of the distributions of our metrics:
```
EmilHvitfeldt:
```suggestion
Now that we've evaluated a number of models with a variety of metrics, we can explore the results to determine our optimal model. Beginning with a quick exploratory plot of the distributions of our metrics:
```

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,516 @@
...
+
+```{r}
+#| label: "autoplot-age-bt"
+#| out.width: 70%
+#| fig.width: 6
+#| fig.height: 7
+autoplot(wflow_set_fit, id = "age_bt")
+```
+
+The learning rate `learn_rate` seems to have a more pronounced effect on the results metrics than the number of randomly selected predictors `mtry`. As before, we see that the most performant models with respect to `roc_auc()` tend to be the most fair with respect to our fairness metrics. Further, the values of each of the fairness metrics plotted above seem highly correlated.
```
EmilHvitfeldt: Minor, is it confusing that the words "with respect to" are used quite close to eachother?

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,516 @@
...
+```
+
+The learning rate `learn_rate` seems to have a more pronounced effect on the results metrics than the number of randomly selected predictors `mtry`. As before, we see that the most performant models with respect to `roc_auc()` tend to be the most fair with respect to our fairness metrics. Further, the values of each of the fairness metrics plotted above seem highly correlated. 
+
+From the perspective of a practitioner hoping to satisfy various stakeholders, the fact that these metrics are highly correlated makes the model selection process much easier. We can choose one fairness metric that we'd like to optimize for, and likely end up with a near-optimal configuration for the other metrics as a byproduct.
+
+:::callout-note
+In machine learning fairness, "impossibility thereoms" show that fairness definitions "are not mathematically or morally compatible in general" [@mitchell2021]. More concretely, unless we live in a world with no inequality, there is no way to satisfy many definitions of fairness at once. However, recent research emphasizes that near-fairness among more limited sets of metrics, like the three we've used here, is both possible and relevant [@bell2023].
```
EmilHvitfeldt:
```suggestion
In machine learning fairness, "impossibility theorems" show that fairness definitions "are not mathematically or morally compatible in general" [@mitchell2021]. More concretely, unless we live in a world with no inequality, there is no way to satisfy many definitions of fairness at once. However, recent research emphasizes that near-fairness among more limited sets of metrics, like the three we've used here, is both possible and relevant [@bell2023].
```

EmilHvitfeldt (approved):
A few notes.

It looks amazing, so excited to see this series of posts coming together!

learn/work/fairness-readmission/index.qmd
simonpcouch: Ah, because there's a typo!😄 Thank you. [e21af27](https://github.com/tidymodels/tidymodels.org/pull/52/commits/e21af27668c751d5159b7f0539c321000860814f).

learn/work/fairness-readmission/index.qmd
simonpcouch: [e21af27](https://github.com/tidymodels/tidymodels.org/pull/52/commits/e21af27668c751d5159b7f0539c321000860814f).

learn/work/fairness-readmission/index.qmd
simonpcouch: [e21af27](https://github.com/tidymodels/tidymodels.org/pull/52/commits/e21af27668c751d5159b7f0539c321000860814f).

learn/work/fairness-readmission/index.qmd
simonpcouch: [e21af27](https://github.com/tidymodels/tidymodels.org/pull/52/commits/e21af27668c751d5159b7f0539c321000860814f).

learn/work/fairness-readmission/index.qmd
simonpcouch: [e21af27](https://github.com/tidymodels/tidymodels.org/pull/52/commits/e21af27668c751d5159b7f0539c321000860814f).

learn/work/fairness-readmission/index.qmd
```diff
@@ -0,0 +1,515 @@
...
+
+A vast majority of patients are labeled `"Caucasian"` (`r round(mean(readmission$race == "Caucasian"), 3)*100`%) or `"African American"` (`r round(mean(readmission$race == "African American"), 3)*100`%). The counts for the remaining racial categorizations are quite a bit smaller, and when we split the data up into resamples, those counts will reduce even further. As a result, the variability associated with the estimates `"Asian"`, `"Hispanic"`, `"Other"`, and `"Unknown"` will be larger than those for `"African American"` and `"Causasian"`. As an example:
+
+```{r}
+#| label: "prop-by-race"
+#| message: false
+#| warning: false
+readmission %>%
+  # randomly split into groups
+  mutate(., group = sample(1:10, nrow(.), replace = TRUE)) %>%
```
EmilHvitfeldt, line 122: You should put a `set.seed()` before this code

simonpcouch: Good point! [dd025fb](https://github.com/tidymodels/tidymodels.org/pull/52/commits/dd025fbef72a94c47c04159423fc97a8cb14fc54).

PR #61, opened 2024-03-22
Update survival articles post release

hfrick:
This removes the reference to the dev versions of some of the packages. Should I render the two articles and add those changes as well?


EmilHvitfeldt:
sure thing!

PR #68, opened 2024-04-12
initial draft of percentile interval article

topepo:


404.html.md
```diff
@@ -4,6 +4,8 @@ title: Page Not Found
 
 
 
+
```
topepo, line 7: I have been using a very recent quarto daily. Let me know if there is a specific version that I should use.

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+delivery_split <- initial_validation_split(deliveries, prop = c(0.6, 0.2), 
+                                           strata = time_to_delivery)
+
+# Make data frames for each of the three data sets
+delivery_train <- training(delivery_split)
+delivery_test  <- testing(delivery_split)
+delivery_val   <- validation(delivery_split)
+
+# Create an object that makes a validation set "look like resampling"
+# to tidymodels:
```
hfrick: This could also be something along the lines of "bundle training and validation set as a resample" to be explicit about what does into the object. "Making a validation set look like resample" could be misunderstood to mean only the validation set is included in `delivery_rs`.

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+# quietly use parallel process
+library(doMC)
+
+tidymodels_prefer()
+theme_set(theme_bw())
+options(pillar.advice = FALSE, pillar.min_title_chars = Inf)
+registerDoMC(cores = parallel::detectCores(logical = FALSE))
+```
+
+We'll use the [delivery time data](https://modeldata.tidymodels.org/reference/deliveries.html) and follow the analysis used on [this website](https://aml4td.org/chapters/whole-game.html#sec-delivery-times). The outcome is the time for food to be delivered, and the predictors include the day/hour of the order, the distance, and what was included in the order (columns starting with `item_`):
```
hfrick: "this website" is a bit vague. Unless you intentionally do that I would put an explicit reference to the book title.

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+```
+
+Let's use grid search to evaluate a grid of values between 2 and 50. We'll use `tune_grid()` with an option to save the out-of-sample (i.e., validation set) predictions for each candidate model in the grid. By default, for regression models, the function computes the root mean squared error (RMSE) and R<sup>2</sup>:  
+
+```{r}
+#| label: grid
+#| cache: true
+
+grid <- tibble(num_terms = 2:50)
+
```
hfrick:
```suggestion
```

nit :)

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+```
+
+How did the model look?
+
+```{r}
+#| label: perf-plot
+autoplot(mars_res)
+```
+
+After about 20 retained terms, both statistics appear to plateau.  However, we have no sense of the noise around these values. Is a model with 20 terms just as good as a model using 50?
```
hfrick: When I read this, I wonder more if a model with 12 terms is just as good as with 20, wanting a simpler model with the same performance.
```suggestion
After about 20 retained terms, both statistics appear to plateau.  However, we have no sense of the noise around these values. Is a model with 12 terms just as good as a model using 20? 
```

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+
+## Introduction 
+
+The tidymodels framework focuses on evaluating models via _empirical validation_: out-of-sample data are used to compute model accuracy/fitness measures. Because of this, data splitting and resampling are essential components of model development. 
+
+If a model uses traditional resampling (such as 10-fold cross-validation), it is easy to get confidence intervals (or Bayesian intervals) of performances. For example, if you are looking at classification accuracy, you can say something like, 
+
+> Our accuracy was estimated to be 91.3% with a 90% confidence interval of (80.1%, 95.9%).
+
+For some performance metrics, it is very difficult to compute interval estimates of validation sets (or test sets).
```
hfrick: I would rephrase this a little, to connect it a bit more smoothly to the idea of line 29 with the resampling. I'd start with the validation set being a great choice to reduce the number of model fits when enough data are available, but that this also means only a single value for the metric and hence no "free" interval estimate. Some performance metrics allow you to calculate analytical intervals but most don't - so we need a different tool for these situations!

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+
+For that, we'll compute confidence intervals. However, there are no analytical formulas for most performance statistics, so we need a more general method for computing them.
+
+## Bootstrap Confidence Intervals
+
+In statistics, the bootstrap is a resampling method that takes a random sample the same size as the original but samples [with replacement](https://en.wikipedia.org/wiki/Sampling_(statistics)#Replacement_of_selected_units). This means that, in the bootstrap sample, some rows of data are represented multiple times and others not at all. 
+
+There is [some theory](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C7&q=%22bootstrap+confidence+intervals%22&btnG=) that shows that if we recompute a statistic a large number of times using the bootstrap, we can understand its sampling distribution and, from that, compute confidence intervals. A good recent reference on this subject (but the original inventor of the bootstrap) is [_Computer Age Statistical Inference_](https://hastie.su.domains/CASI/) (which includes a book PDF).
```
hfrick:
```suggestion
There is [some theory](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C7&q=%22bootstrap+confidence+intervals%22&btnG=) that shows that if we recompute a statistic a large number of times using the bootstrap, we can understand its sampling distribution and, from that, compute confidence intervals. A good recent reference on this subject by the original inventor of the bootstrap is [_Computer Age Statistical Inference_](https://hastie.su.domains/CASI/) which is available as a free PDF.
```

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+
+For our application, we'll take the validation set predictions for each candidate model in the grid, bootstrap them, and then compute confidence intervals using the percentile method.
```
hfrick: I think it's worth highlighting here that this means we get interval estimates _without_ having to refit the model. We need to draw bootstrap sample of the predictions and re-calculate the metrics but that is typically "cheaper" than refitting the model.

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+    geom_ribbon(aes(ymin = .lower, ymax = .upper), alpha = 1 / 5, fill = "blue") +
+    facet_wrap(~ .metric, scales = "free_y", ncol = 1) +
+    labs(y = NULL, x = "# Retained Terms")
+```
+
+Those are very tight! Maybe there is some credence to using many terms. Let's say that 40 terms seems like a reasonable value for that tuning parameter. It's large, but the out-of-sample statistics indicate that it does not overfit at this point. 
+
+## Test Set Intervals
+
+Suppose the MARS model was the best we could do for these data. We would then fit the model (with 40 terms) on the training set then finally evaluate the test set. tidymodels has an API called `last_fit()` that uses our original data splitting object (`delivery_split`) and the model specification. To get the test set predictions, we can use `collect_metrics()`:
```
hfrick: I'd just use function rather than API here because it's simpler and offers less opportunity to get distracted by "why are they calling this an API?".

```suggestion
Suppose the MARS model was the best we could do for these data. We would then fit the model (with 40 terms) on the training set then finally evaluate the test set. tidymodels has a function called `last_fit()` that uses our original data splitting object (`delivery_split`) and the model specification. To get the test set predictions, we can use `collect_metrics()`: 
```

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+## Test Set Intervals
+
+Suppose the MARS model was the best we could do for these data. We would then fit the model (with 40 terms) on the training set then finally evaluate the test set. tidymodels has an API called `last_fit()` that uses our original data splitting object (`delivery_split`) and the model specification. To get the test set predictions, we can use `collect_metrics()`: 
+
+```{r}
+#| label: test-res
+
+mars_final_spec <- 
+    mars(num_terms = 40, prod_degree = 2, prune_method = "none") %>% 
+  set_mode("regression")
```
hfrick:
```suggestion
mars_final_spec <- 
  mars(num_terms = 40, prod_degree = 2, prune_method = "none") %>% 
  set_mode("regression")
```

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+set.seed(168)
+mars_test_boot <- int_pctl(mars_test_res, alpha = 0.10)
+mars_test_boot
+```
+
+## Notes
+
+ - The point estimates produced by `collect_metrics()` and `int_pctl()` may disagree. They are two different statistical estimates of the same quantity. For reasonably sized data sets, they should be pretty close. 
+ 
+ - Parallel processing can be used for bootstrapping (using the [same tools](https://www.tmwr.org/grid-search#parallel-processing) that can be used with the tune package).
```
hfrick: The tune page has some practical advice on setting those tools up: https://tune.tidymodels.org/articles/extras/optimizations.html#parallel-processing

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+```
+
+## Notes
+
+ - The point estimates produced by `collect_metrics()` and `int_pctl()` may disagree. They are two different statistical estimates of the same quantity. For reasonably sized data sets, they should be pretty close. 
+ 
+ - Parallel processing can be used for bootstrapping (using the [same tools](https://www.tmwr.org/grid-search#parallel-processing) that can be used with the tune package). 
+ 
+ 
+ - `int_pctl()` works with classification and censored regression models. For the latter, the default is to compute intervals for every metric and every evaluation time.
```
hfrick:
```suggestion
 - `int_pctl()` also works with classification and censored regression models. For the latter, the default is to compute intervals for every metric and every evaluation time. 
```

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+
+`int_pctl()` also works on objects produced by `last_fit()`: 
+
+```{r}
+#| label: test-boot
+set.seed(168)
+mars_test_boot <- int_pctl(mars_test_res, alpha = 0.10)
+mars_test_boot
+```
+
```
hfrick, line 190: I would add a summary sentence or two on the main point here, just to round things out before the notes. To me, those main points are:
```suggestion

So, to sum up the main idea: If you're not getting multiple estimates of your performance metric from your resample procedure—like when using a validation set—you can still get interval estimates for your metrics. A metric-agnostic approach is to bootstrap your predictions and recalculate your metrics based on those.

```

learn/models/bootstrap-metrics/index.qmd
hfrick: Having read further, I see that you were very intentional about the higher number of terms. Maybe put the motivation for which model you're interested in up here instead of where it is now. Still wondering why you'd want more terms though :)

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+mars_boot %>% 
+  ggplot(aes(num_terms)) + 
+    geom_line(aes(y = .estimate)) + 
+    geom_point(aes(y = .estimate), cex = 1 / 2) + 
+    geom_ribbon(aes(ymin = .lower, ymax = .upper), alpha = 1 / 5, fill = "blue") +
+    facet_wrap(~ .metric, scales = "free_y", ncol = 1) +
+    labs(y = NULL, x = "# Retained Terms")
+```
+
+Those are very tight! Maybe there is some credence to using many terms. Let's say that 40 terms seems like a reasonable value for that tuning parameter. It's large, but the out-of-sample statistics indicate that it does not overfit at this point.
```
hfrick: Ah!

hfrick (approved):
General structure comment: I'm wondering if you'd want to put the idea first, since there is not actually that much detail, and then follow up with the example. Right now, there is some motivation at the start and then some explanation on the bootstrap later but I think it might work to pull both parts up front. That would make them more visible, and easier to use this post as a reference/refresher.

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
+---
+title: "Confidence Intervals for Performance Metrics"
+categories:
+  - model fitting
+  - confidence intervals
+  - bootstrapping
+type: learn-subsection
+weight: 6
+description: | 
+  tidymodels has high-level functions to produnce bootstrap confidence intervals for performace statistics.
```
EmilHvitfeldt:
```suggestion
  tidymodels has high-level functions to produce bootstrap confidence intervals for performance statistics.
```

learn/models/bootstrap-metrics/index.qmd
```diff
@@ -0,0 +1,212 @@
...
+ - The point estimates produced by `collect_metrics()` and `int_pctl()` may disagree. They are two different statistical estimates of the same quantity. For reasonably sized data sets, they should be pretty close. 
+ 
+ - Parallel processing can be used for bootstrapping (using the [same tools](https://www.tmwr.org/grid-search#parallel-processing) that can be used with the tune package). 
+ 
+ 
+ - `int_pctl()` works with classification and censored regression models. For the latter, the default is to compute intervals for every metric and every evaluation time. 
+
+ - `int_pctl()` has options to filter which models, metrics, and/or evaluation times are used for analysis. If you investigated many grid points, you don't have to bootstrap them all. 
+ 
+ - Since bootstrap percentiles use... percentiles, you should generate a few thousand bootstrap samples to compute the intervals to get stable and accurate results.
```
EmilHvitfeldt: the dots here feels a little too informal for this type of content

EmilHvitfeldt (approved):
✨

PR #72, opened 2024-06-25
consolidate tag `categories`

simonpcouch:
My goal here was to make "categories" in the Learn listing more navigable by halving the number of entries. I mostly did so by 1) removing categories that only have one listing (especially if the category is in the title of the article) and 2) aligning synonyms for the same category, like "tune," "tuning," and "model tuning." 

I generally didn't prefer package names over keywords or vice versa—that may be a discussion worth having for a future PR.


simonpcouch:
Expected to likely see an error before rendering, but was surprised to see this specifically:

```
ERROR: Error executing 'Rscript': No such file or directory (os error 2)

Unable to locate an installed version of R.
```

Any intuition here before I start debugging, @EmilHvitfeldt?

EmilHvitfeldt:
Appears to be this issue: https://github.com/quarto-dev/netlify-plugin-quarto/issues/2

EmilHvitfeldt:
but if you look at the netlify logs, it doesn't break right away, the first breakage is at `learn/models/calibration`, so to fix this "easy" we just have to update the `_freeze` for the breaking pages/the pages you updated

EmilHvitfeldt:
works now 🎉

PR #76, opened 2024-08-13
size > linewidth

abichat:


EmilHvitfeldt:
Thanks for finding this mistake!

PR #85, opened 2024-11-27
Update article on how to make a parsnip model

hfrick:
A few changes happened in the past two years:
- We have a new mode: Hello censored regression!
- We've settled on an API for `multi_predict()`.
- We are a bit smarter about quosures.

Otherwise, I've just made the styling more consistent.


hfrick:
The actions failed when I didn't render, so I did render. But now I'm questioning if that was a good decision since it also rendered some other articles. (I did remember to use CRAN versions though.)

learn/develop/models/index.qmd
```diff
@@ -47,22 +47,22 @@ This article describes the process of creating a new model function. Before proc
...
 ```
 
-The main hyperparameter is the number of subclasses. We'll name our function `discrim_mixture`. 
+The main hyperparameter is the number of subclasses. We'll name our function `discrim_mixture()`. 
 
 ## Aspects of models
 
 Before proceeding, it helps to to review how parsnip categorizes models:
 
 * The model _type_ is related to the structural aspect of the model. For example, the model type `linear_reg` represents linear models (slopes and intercepts) that model a numeric outcome. Other model types in the package are `nearest_neighbor`, `decision_tree`, and so on.
```
EmilHvitfeldt: do we wanna add `()` to these too since they are function calls.

`linear_reg` -> `linear_reg()`?

learn/develop/models/index.qmd
```diff
@@ -139,7 +139,7 @@ discrim_mixture <-
   function(mode = "classification",  sub_classes = NULL) {
     # Check for correct mode
     if (mode  != "classification") {
-      rlang::abort("`mode` should be 'classification'")
+      rlang::abort("`mode` should be 'classification'.")
```
EmilHvitfeldt, line 142: should we encourage {cli} here?

learn/develop/models/index.qmd
```diff
@@ -160,14 +160,14 @@ discrim_mixture <-
 This is pretty simple since the data are not exposed to this function. 
 
 ::: {.callout-warning}
- We strongly suggest favoring `rlang::abort()` and `rlang::warn()` over `stop()` and `warning()`. The former return better traceback results and have safer defaults for handling call objects. 
+ We strongly suggest favoring `rlang::abort()` and `rlang::warn()` (or their cli counterparts `cli::cli_abort()` and `cli::cli_warn()`) over `stop()` and `warning()`. The former return better traceback results and have safer defaults for handling call objects.
```
EmilHvitfeldt, line 163: I guess we kinda do 😄

learn/develop/models/index.qmd
```diff
@@ -516,26 +516,25 @@ show_best(mda_tune_res, metric = "roc_auc")
 
 There are various things that came to mind while developing this resource.
 
-**Do I have to return a simple vector for `predict` and `predict_class`?**
+**Do I have to return a simple vector for `predict()`?**
 
-Previously, when discussing the `pred` information:
+There are some models (e.g. `glmnet`, `plsr`, `Cubist`, etc.) that can make predictions for different models from the same fitted model object. We facilitate this via `multi_predict()`, rather than `predict()`.
```
EmilHvitfeldt: Are `glmnet`, `plsr`, `Cubist` referring to packages or engines? if engines they should be in quotes

EmilHvitfeldt (commented):
i like it.

i wouldn't worry too much about the other files. it appears there wasn't that many changes

EmilHvitfeldt (approved):
i like it.

i wouldn't worry too much about the other files. it appears there wasn't that many changes

learn/develop/models/index.qmd
hfrick: That was my first impulse as well. I was on the fence if the text was not intentionally distinguishing between the type and the model function but if you read it that way as well, I'll update!

learn/develop/models/index.qmd
hfrick: I didn't want to add cli styling here because I think it might not be that common yet that it's easy to read for most people - and then it just distracts from the main point. Instead, I added a reference to cli in the callout box.

learn/develop/models/index.qmd
hfrick: These are not the engine names. I'm gonna remove the code styling instead (assuming that the general names are better known than the engine names.)

PR #86, opened 2024-11-29
Grid processing article

topepo:
An incomplete first pass for an article outlining how we process grids inside of `tune_grid()`. One purpose of this is to come up with model code to encapsulate the different stages of training, prediction, and adjustment. 

The text is a little cheeky, reflecting my stream-of-consciousness writing style for drafts (but that;s why it is currently a draft PR).


topepo:
I'm going to rewrite this from scratch. I'll close and start a fresh PR later.

PR #95, opened 2025-03-12
Sparse find page

EmilHvitfeldt:
This PR adds a page under the "Find" section of the website. This page list all model x engine combinations that support sparse data, as well as all the steps that can generate and handle sparse data


find/sparse/index.qmd
```diff
@@ -0,0 +1,75 @@
...
+library(DT)
+tidymodels_prefer()
+models <- read_csv("models.csv", show_col_types = FALSE)
+steps_generate <- read_csv("steps_generate.csv", show_col_types = FALSE)
+steps_preserve <- read_csv("steps_preserve.csv", show_col_types = FALSE)
+```
+
+## Models
+
+Below is all the model engine combinations that support sparse data. It is known that `ranger` supports sparse data as an input, it doesn’t use it any differently than dense data. Thus it shouldn't be used with truly sparse data.
```
topepo: > Thus it shouldn't be used with truly sparse data.

Maybe say that there isn't much benefit to using sparse structures.

find/sparse/index.qmd
```diff
@@ -0,0 +1,75 @@
...
+}
+
+tags$ul(
+  map(model_split, print_one_model)
+)
+```
+
+## Steps
+
+Sparse data compatibility for steps comes in 2 flavors. The first kind generates sparse data from dense data. Often converting categorical variables to many sparse columns. This type of step is listed here:
```
topepo, line 57: 💯

topepo (approved):
Looks good. Two notes: 

- I don't remember how crucial namespacing is for `make_function_lists.R` but I added more.
- The code wants to get the viruslearner package but that was archived. You might want to make an exclusion list and include that.

PR #111, opened 2025-05-27
Corrections on Brier score for survival predictions

### LINKED ISSUE #83: Formula for time-dependent Brier score on https://www.tidymodels.org/learn/statistics/survival-metrics/
The formula for the time-dependent Brier score on https://www.tidymodels.org/learn/statistics/survival-metrics/ appears to be written incorrectly. In the block for non-events, $y_{it}$ should be replaced by the indicator for a non-event; i.e.,  $1 - y_{it}$. (If this weren't done the numerator for both events and non-events would be the same.) Second, the sum should be divided by $n$ = the number of individuals with non-missing survival data (including censored entries) instead of $W_t$. These are in line with the Graf, 1999 articles and what seems to be coded at https://github.com/tidymodels/yardstick/blob/main/R/surv-brier_survival.R .

hfrick:
closes #83


EmilHvitfeldt (approved):
looks good!

PR #117, opened 2025-08-12
Add a "Create your own score class object" page for filtro 

franceslinyc:
This PR adds a new post titled "Create your own score class object" for filtro in developer tools.

- [x] Finalize 
- [x] Have Hannah review once finalizing
- [x] Revise based on Hannah's feedback
- [x] Ready to merge


franceslinyc:
```
Error:
! Can't find method for `fit(<class_score_aov>)`.
Hide Traceback
    ▆
 1. ├─filtro::fit(score_aov_pval, Sale_Price ~ ., data = ames_subset)
 2. │ └─S7::S7_dispatch()
 3. └─S7:::method_lookup_error("fit", `<named list>`)
```

I think this is related to PR [#162](https://github.com/tidymodels/filtro/pull/162) in filtro

franceslinyc:
Do users need to install `S7`?

franceslinyc:
@topepo @EmilHvitfeldt  Just want to know if there is anything that is missing from the page. 

Still trying to see why `fit()` throws error here. It won't let me Preview nor Render so I commented it out currently.

franceslinyc:
```
Error in `UseMethod()`:
! no applicable method for 'fit' applied to an object of class "c('class_score_aov', 'filtro::class_score', 'S7_object')"
Hide Traceback
    ▆
 1. └─generics::fit(score_aov_pval, Sale_Price ~ ., data = ames_subset)
```

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,284 @@
...
+-   `results`: A slot for the results once the method is fitted. Initially, this is an empty data frame.
+
+## Scoring object specific to the scoring method
+
+As an example, let’s consider the ANOVA F-test filter. 
+
+`class_score_aov` is a subclass of `class_score`. Any additional properties specific to the implementation can be added in the subclass. For example: 
+
+```{r}
+# Create a subclass named 'class_score_aov'
```
EmilHvitfeldt, line 110:
```suggestion
#| eval: false
# Create a subclass named 'class_score_aov'
```

this chunk technically overwrites the same object the package provides. You are thus missing the `S7::method(fit, class_score_aov)` call that adds the fit method. I would advice that we simple show the code without running it,

EmilHvitfeldt:
> Do users need to install `S7`?

Nope, it should be installed as they install {filtro}, as it is in Imports

learn/develop/filtro/index.qmd
franceslinyc: Got you, thanks!

franceslinyc: This solves the issue above, thanks!

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+library(modeldata)
+
+pkgs <- c("filtro", "modeldata")
+```
+
+## Introduction
+
+`r article_req_pkgs(pkgs)`
+
+You can construct new scoring objects using `class_score()`. This article is a guide to creating new scoring objects.
```
hfrick: Can you give readers some context for scoring objects, e.g., via a brief intro to filtro? See the other articles in the "developer tools" category for examples: https://www.tidymodels.org/learn/#category=developer%20tools

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+
+## Introduction
+
+`r article_req_pkgs(pkgs)`
+
+You can construct new scoring objects using `class_score()`. This article is a guide to creating new scoring objects. 
+
+## Scoring object
+
+All subclasses specific to the scoring method have a parent class named `class_score`. There are a few properties (attributes) for this object:
```
hfrick: Can you add a little bit more context on the package structure?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+
+```{r}
+#| label: "class_score"
+library(filtro)
+args(class_score)
+```
+
+-   `outcome_type`: What types of outcome can the method handle? The options are `numeric`, `factor`, or both.
```
hfrick: This is a pretty long list. Does the reader need to know about all of them, at this point? Or could you highlight some and point them to the docs for the rest?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+
+-   `label`: What label to use when printing? For example, `ANOVA p-values`, `ANOVA F-statistics`.
+
+-   `packages`: What packages, if any, are required to train the method?
+
+-   `results`: A slot for the results once the method is fitted. Initially, this is an empty data frame.
+
+## Scoring object specific to the scoring method
+
+We demonstrate how to create a custom scoring object specific to the given scoring method.
```
hfrick: Does this mean that a scoring object is the implementation of a scoring method? The use of "specific to" is throwing me a little here.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+
+As an example, let’s consider the ANOVA F-test filter.
```
hfrick: A sentence on the stats aspect of this would be a nice bonus.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+    fallback_value = Inf,
+    score_type = "aov_fstat",
+    direction = "maximize",
+    deterministic = TRUE,
+    tuning = FALSE,
+    label = "ANOVA F-statistics"
+  )
+```
+
+## Fitting (or estimating) score
```
hfrick: Does this use the fit method for the parent class? I was expecting to see some class-specific method implementations in this section.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,289 @@
...
+  fit(Sale_Price ~ ., data = ames)
+score_cor_spearman |>
+  fit(Sale_Price ~ ., data = ames)
+```
+
+## Documenting S7 methods 
+
+Documentation for S7 methods is still a work in progress, and it seems no one currently knows the right approach. Here’s how we tackle it: 
+
+We re-export the `fit()` generic from another package. Instead of documenting each individual `fit()` method, we provide the details in the "Details" section and the "Estimating the scores" subsection of the documentation for the corresponding object.
```
hfrick: I would show examples of that. The call to the help pages below shows users how to access it and that's helpful context but it less focused on "how to implement your own scoring class".

hfrick (commented):
Looks like you made a very cool package with filtro! I would add a little bit more context in the article. 

You don't have to write a novel about what filtro does (that's what the filtro website is for) but just a few sentences to situate a reader who's possibly coming from the developer category on tidymodels.org and not directly from the filtro page. 

I'd also add some broad strokes about how filtro works as context for the concrete steps of what to actually do when implementing a new scoring method. And then always show concrete examples, like for the docs.

franceslinyc:
All super important points, thanks @hfrick! I'll revise according.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+pkgs <- c("filtro", "modeldata")
+```
+
+## Introduction
+
+`r article_req_pkgs(pkgs)`
+
+filtro is tidy tools to apply filter-based supervised feature selection methods. It provides functions to rank and select a specified proportion or a fixed number of features using built-in methods and the desirability function. 
+
+Currently, there are 6 filters in filtro and many existing score objects. A list of existing scoring objects [can be found here](https://filtro.tidymodels.org/articles/filtro.html#available-score-objects-and-filter-methods). However, you might need to define your own scoring objects. This article serves as a guide to creating new scoring objects and computing feature scores before performing ranking and selection.
```
hfrick, line 37: What's the purpose of a scoring object? So far, the text only mentions filters.

hfrick: Ah, some explanation coming at the end of the paragraph. Could we reshuffle that a little?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+
+`r article_req_pkgs(pkgs)`
+
+filtro is tidy tools to apply filter-based supervised feature selection methods. It provides functions to rank and select a specified proportion or a fixed number of features using built-in methods and the desirability function. 
+
+Currently, there are 6 filters in filtro and many existing score objects. A list of existing scoring objects [can be found here](https://filtro.tidymodels.org/articles/filtro.html#available-score-objects-and-filter-methods). However, you might need to define your own scoring objects. This article serves as a guide to creating new scoring objects and computing feature scores before performing ranking and selection. 
+
+The general procedure is to:
+
+1. Create a parent scoring object `class_score`, specifying fixed properties that are shared across all custom scoring objects.
```
hfrick: This class is already implemented in filtro, right? If so, we should make it clearer that this is the infrastructure we build on when we make our own score class object.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+
+filtro is tidy tools to apply filter-based supervised feature selection methods. It provides functions to rank and select a specified proportion or a fixed number of features using built-in methods and the desirability function. 
+
+Currently, there are 6 filters in filtro and many existing score objects. A list of existing scoring objects [can be found here](https://filtro.tidymodels.org/articles/filtro.html#available-score-objects-and-filter-methods). However, you might need to define your own scoring objects. This article serves as a guide to creating new scoring objects and computing feature scores before performing ranking and selection. 
+
+The general procedure is to:
+
+1. Create a parent scoring object `class_score`, specifying fixed properties that are shared across all custom scoring objects. 
+
+2. Construct a custom scoring object `class_score_*`, adding additional, implementation-specific properties.
```
hfrick: I'd rephrase this to say that we are making a _subclass_, inheriting from `class_score`. The info that this is the place to specify aspects specific to this score (as opposed to the general characteristics already covered by the parent class) is proabaly helpful for orientating readers less familar with object-oriented programming.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+
+Currently, there are 6 filters in filtro and many existing score objects. A list of existing scoring objects [can be found here](https://filtro.tidymodels.org/articles/filtro.html#available-score-objects-and-filter-methods). However, you might need to define your own scoring objects. This article serves as a guide to creating new scoring objects and computing feature scores before performing ranking and selection. 
+
+The general procedure is to:
+
+1. Create a parent scoring object `class_score`, specifying fixed properties that are shared across all custom scoring objects. 
+
+2. Construct a custom scoring object `class_score_*`, adding additional, implementation-specific properties. 
+
+3. Define the scoring method in `fit()`, which computes feature score. `fit()` refers to the custom scoring object from step 2 to use the appropriate method.
```
hfrick: If you think it's helpful, you could clarify that defining the subclass corresponds to defining characteristics of your score and writing the fit() method corresponds to writing down how to calculate it.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+
+-   `case_weights`: Does the method accpet case weights? It is `TRUE` or `FALSE`.
+
+-   `fallback_value`: What is a value that can be used for the statistic so that it will never be eliminated? For example, `0`, `Inf`.
+
+-   `direction`: What direction of values indicates the most important values? For example,  `maximum`, `minimize`.
+
+-   `results`: A slot for the results once the method is fitted. Initially, this is an empty data frame.
+
+For details on its constructor and its remaining properties, please refer to the package documentation.
```
hfrick: It'b be nice to have a link here directly to the relevant docs.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+
+## Custom scoring object
+
+```
+class_score
+└─> class_score_aov (example shown)
+└─> class_score_cor
+└─> ... 
+```
+
```
hfrick, line 101: This is sort of a bridge between the previous section and this section. That is super helpful but I would spell this out, along the lines of "all scores implemented in filtro are subclasses of this base class which means they inherit from this base class. so when we create a new score, we create another subclass of this." Such a diagram should come after some explanation but on it's own it's dangling a bit. I could see this at the end of the previous section or the start of this section, whereever you think it fits best.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+Next, we demonstrate how to create a custom scoring object `class_score_*`. 
+
+As an example, let’s create a custom scoring object for ANOVA F-test named `class_score_aov`. This filter computes feature score using analysis of variance (ANOVA) hypothesis tests, powered by `lm()`. The `lm()` function fits a linear model and returns a summary containing the F-statistic and p-value, which can be used to evaluate feature importance.
```
hfrick: Nice! Thank you for the concise explanation here.

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+library(modeldata)
+
+pkgs <- c("filtro", "modeldata")
+```
+
+## Introduction
+
+`r article_req_pkgs(pkgs)`
+
+filtro is tidy tools to apply filter-based supervised feature selection methods. It provides functions to rank and select a specified proportion or a fixed number of features using built-in methods and the desirability function.
```
hfrick: yes! that helps setting the scene

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+args(class_score)
+```
+
+For example: 
+
+-   `case_weights`: Does the method accpet case weights? It is `TRUE` or `FALSE`.
+
+-   `fallback_value`: What is a value that can be used for the statistic so that it will never be eliminated? For example, `0`, `Inf`.
+
+-   `direction`: What direction of values indicates the most important values? For example,  `maximum`, `minimize`.
```
hfrick: Should this be `maximum` and `minimum` or `maximize` and `minimize`?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+
+Individual properties can be accessed via `object@`. For example: 
+
+```{r}
+score_aov_pval@case_weights
+score_aov_pval@fallback_value
+score_aov_pval@direction
+```
+
+Note that by default, the returned p-value is transformed to `-log10(p_value)`, which means larger values correspond to more important predictors. This is why the fallback value is set to `Inf` and the direction is set to `"maximize"`.
```
hfrick, line 161: That note is so helpful, I was giving that "maximize" a raised eyebrow already!

hfrick: How does that play out in the class variant for the F-statistic?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+#| eval: false
+# Define the scoring method for `class_score_aov`
+S7::method(fit, class_score_aov) <- function(
+  object,
+  formula,
+  data,
+  case_weights = NULL,
+  ...
+) {
+  # TODO Finish the rest of the function
```
hfrick: Is this meant as a TODO for you now or do you plan to leave it in for the reader?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+  object@results <- res
+  object
+}
+```
+
+We would want to do something similar for other `class_score_*` subclass. 
+
+## Documenting S7 methods 
+
+Documentation for S7 methods is still a work in progress, and it seems no one currently knows the right approach. Here’s how we tackle it:
```
hfrick: Maybe something along the lines of "Documentation for S7 methods is still a work in progress, but our current best approach is as follows"?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+  case_weights = NULL,
+  ...
+) {
+  # TODO Finish the rest of the function using lm() and anova()
+
+  object@results <- res
+  object
+}
+```
+
```
hfrick: I know I asked for examples previously but I'm leaning towards removing this one here because it's just one line that's relevant here. I think we can capture that in the text above. What do you think?

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+Instead, documentation is provided in the "Details" section and the "Estimating the scores" subsection of the documentation for the `score_aov_pval` object.
```
hfrick, line 292: The example for this one I find super helpful!

learn/develop/filtro/index.qmd
```diff
@@ -0,0 +1,418 @@
...
+```{r}
+ames_aov_fstat_res@results
+```
+
+## Session information {#session-info}
+
+```{r}
+#| label: "si"
+#| echo: false
+small_session(pkgs)
```
hfrick, line 420: For the final rendering, could you update filtro to the latest CRAN version? We have a convention of using only CRAN versions for these articles :)

hfrick (commented):
Thank you for the added context! I have a much better orientation in this now. 🐸

My understanding is that the core things to know about adding your own scoring method are:

There is a base class for scoring objects. You add a scoring method by
1. Creating a subclass of that base class
2. Writing a `fit()` method for the subclass

And for the docs, you get guidance on how to document an S7 method.

Assuming I got that correct, I think the article could make that just a little clearer and then we're there!

learn/develop/filtro/index.qmd
franceslinyc: Just saw these now, adding (revising)! Thanks!

franceslinyc:
Still working!

learn/develop/filtro/index.qmd
franceslinyc: Yes, I quoted Max too directly. 😅

learn/develop/filtro/index.qmd
franceslinyc: I truncated it down further, but I think it may still good to include the skeleton code here? Point of view from someone who is still navigating roxygen.

franceslinyc:
Thanks so much for the feedback @hfrick, I hope I didn't miss anything. @topepo or @EmilHvitfeldt, would you like to take a final look?

PR #119, opened 2025-09-16
rephrase link to model engine table (that doesn't show args)

corybrunson:
This is a one-line change to instruct readers to view topic pages linked from [the model/engine table](https://www.tidymodels.org/find/parsnip/), since the current table does not show, and cannot be searched by, arguments.

Apologies if i've misunderstood something! Happy to revise if requested.


EmilHvitfeldt:
Thank you for the clarification!

PR #122, opened 2025-09-19
Added CSS for Resources sidebar to override Quarto default sidebar CSS

dsluder13:
When Quarto renders a website, files are passed through a file that apples logic to determine which CSS to assign different elements. Sidebars can have the [zindex-bottom](https://github.com/quarto-dev/quarto-cli/blob/6eb757f663dae7f3992a1ad502ec730cf0d544b9/src/format/html/format-html-bootstrap.ts#L1089) class applied to it, and this happens for the Resources sidebar. This [CSS class](https://github.com/quarto-dev/quarto-cli/blob/6eb757f663dae7f3992a1ad502ec730cf0d544b9/src/resources/formats/html/_quarto-rules.scss#L29) causes the element to always be under everything else, which allows the user to view the text but they are unable to interact with it, which means they cannot click on links.

This change adds more specific CSS to override this behavior and allow the user to click on the links.


EmilHvitfeldt:
Thank you!

PR #123, opened 2025-09-19
Set seed for bootsrap examples

### LINKED ISSUE #65: make sure we set seeds everywhere
They don't look problematic! It looks like we don't set the seed where we probably ought to (before `generate()`ing bootstraps), though, which could be a TODO.

_Originally posted by @simonpcouch in https://github.com/tidymodels/tidymodels.org/pull/64#discussion_r1539838004_

aluxh:
Set seed for all the bootstrap examples, but didn't re-render the quarto file. Hope this is alright. Fixes #65


topepo (approved):

topepo:
Thanks!

PR #125, opened 2025-10-13
Article on automatic calibration splits

hfrick:
I've written about how we make the automatic calibration splits. The article covers the guiding principles of our approach, both for how and why (although not in academic paper depth). 

The goal is also to let people understand in detail what happens for the sliding resamples. It has gotten relatively lengthy, though. I'm wondering if it should stay here or, e.g., go into a separate article, similar to how we split out the details on how we deal with censoring for the dynamic survival metrics. I think there's value in working through those details somewhere (other than the source code directly), but we could also experiment with collapsible text. Do you have any preferences or suggestions? Or do you think the length is fine as it is?


learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+categories:
+  - post-processing
+  - calibration
+  - rsample
+type: learn-subsection
+weight: 6
+description: | 
+  Learn how tidymodels generates automatic calibration sets for post-processing.
+toc: true
+toc-depth: 2
```
EmilHvitfeldt, line 12:
```suggestion
toc-depth: 3
```

due to the length of the article i think upping the TOC depth is nice, otherwise we don't update the toc that fast

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+theme_set(theme_bw() + theme(legend.position = "top"))
+```
+
+While preprocessing is the transformation of the predictors prior to a model fit, post-processing is the transformation of the predictions after the model fit. This could be as straightforward as limiting predictions to a certain range of values to as complicated as transforming them based on a separate calibration model. 
+
+A calibration model is used to model the relationship between the predictions based on the primary model and the true outcomes. An additional model means an additional chance to accidentially overfit. So when working with calibration, this is crucial: we cannot use the same data to fit our calibration model as we use to assess the combination of primary and calibration model. Using the same data to fit the primary model and the calibration model means the predictions used to fit the calibration model are re-predictions of the same observations used to fit the primary model. Hence they are closer to the true values than predictions on new data would be and the calibration model doesn't have accurate information to estimate the right trends (so that they then can be removed).
```
EmilHvitfeldt, line 38:
```suggestion
A calibration model is used to model the relationship between the predictions based on the primary model and the true outcomes. An additional model means an additional chance to Accidentally overfit. So when working with calibration, this is crucial: we cannot use the same data to fit our calibration model as we use to assess the combination of primary and calibration model. Using the same data to fit the primary model and the calibration model means the predictions used to fit the calibration model are re-predictions of the same observations used to fit the primary model. Hence they are closer to the true values than predictions on new data would be and the calibration model doesn't have accurate information to estimate the right trends (so that they then can be removed).
```

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+#| warning: false
+library(tidymodels)
+
+pkgs <- c("tidymodels", "rsample")
+
+theme_set(theme_bw() + theme(legend.position = "top"))
+```
+
+While preprocessing is the transformation of the predictors prior to a model fit, post-processing is the transformation of the predictions after the model fit. This could be as straightforward as limiting predictions to a certain range of values to as complicated as transforming them based on a separate calibration model. 
+
```
EmilHvitfeldt, line 37: Below we are using the term **primary model** which [we just started using](https://github.com/search?q=org:tidymodels++%22primary+model%22&type=code). I like the term but i think it be nice to properly define it, in terms of pre/model/post diagram/terminology

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+A calibration model is used to model the relationship between the predictions based on the primary model and the true outcomes. An additional model means an additional chance to accidentially overfit. So when working with calibration, this is crucial: we cannot use the same data to fit our calibration model as we use to assess the combination of primary and calibration model. Using the same data to fit the primary model and the calibration model means the predictions used to fit the calibration model are re-predictions of the same observations used to fit the primary model. Hence they are closer to the true values than predictions on new data would be and the calibration model doesn't have accurate information to estimate the right trends (so that they then can be removed).
+
+rsample provides a collection of functions to make resamples for empirical validation of prediction models. So far, the assumption was that the prediction model is the only model that needs fitting, i.e., a resample consists of an analysis set and an assessment set.
```
EmilHvitfeldt, line 40: Do we have a rsample doc page for analysis set and assessment set?

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+If we include calibration into our workflow (bundeling preprocessing, (primary) model, and post-processing), we want an analysis set, a calibration set, and an assessment set.
```
EmilHvitfeldt, line 42:
```suggestion
If we include calibration into our workflow (bundling preprocessing, (primary) model, and post-processing), we want an analysis set, a calibration set, and an assessment set. 
```

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+"Taking data from the analysis set" means splitting up the analysis set to end up with ... an analysis set and a calibration set. Now we have two sets called analysis set, that's confusing. If we need to distinguish them, we'll refer to them as "outer" and "inner" analysis set for "before" and "after" the split for a calibration set.
+
+
+```{r}
+#| label: calibration
+#| echo: false
+#| out.height: 350
+#| fig.align: "center"
+knitr::include_graphics("images/analysis-calibration-assessment.jpg")
```
EmilHvitfeldt, line 60: How are these images created? if there are source files I would like for them to be included in this folder. Otherwise they will be hard to tweak or reproduce later on

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+rsample provides a collection of functions to make resamples for empirical validation of prediction models. So far, the assumption was that the prediction model is the only model that needs fitting, i.e., a resample consists of an analysis set and an assessment set.
+
+If we include calibration into our workflow (bundeling preprocessing, (primary) model, and post-processing), we want an analysis set, a calibration set, and an assessment set. 
+
+Here we describe how we create these sets automatically when you resample or tune a workflow with calibration. Note that you currently cannot create manual calibration splits and supply them to the functions in tune, like `fit_resamples()` or `tune_grid()`. If you want to do this, leave us a [feature request](https://github.com/tidymodels/rsample/issues/new?template=feature_request.md).
+
+## General principles
+
+First principle: we leave the assessment set purely for performance assessment and rather treat the analysis set as "the data to fit" - except now we split it to fit two different models instead of just one. Or in other words, we take the data for the calibration set from the analysis set, not from the assessment set.
```
EmilHvitfeldt, line 48: I'm writing most of these as I read, so disregard this one if it is answered later and i forget to remove this one.

Should we explain why we decide to split the analysis set instead of the assessment set. or at least mention that we will go over why later?

EmilHvitfeldt: and nit pick, i think using some highlighting on "first principle" and "second principle" would be nice.

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+And lastly, the third principle: if we cannot make a calibration split, we fall back onto an empty calibration set rather than erroring.
+
+Let's look at v-fold cross-validation as an example.
+
+
+### Example: v-fold cross-validation
+
+For v-fold cross-validation, we split the data into `v` folds and create `v` resamples by always using one of the folds as the assessment set while combining the other folds into the analysis set. 
+
+Since we want to have a calibration split similar in spirit, we are splitting the (outer) analysis set into (inner) analysis set and calibration set with the same proportion of observations going into the analysis set: 1 - 1/v.
```
EmilHvitfeldt, line 76: 1 - 1/v should either be styled as code or as latex, but don't care which on

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+- `sliding_index()` works in much the same way, with the key difference being that what we are sliding over aren't the rows directly but rather an index. This is useful when dealing with irregular series of data: while the number of rows which fall into a window can vary, the window length is the same across all resamples.
+- `sliding_period()` takes the idea of the `sliding_index()` function one step further and aggregates the index into periods before sliding across. This is useful for aggregating, e.g., daily data into months and then defining the analysis and assessment set in terms of months rather than the daily index directly.
+
+### Row-based splitting
+
+Let's start with the row-based splitting done by `sliding_window()`. We'll use a very small example dataset. This will make it easier to illustrate how the different subsets of the data are created but note that it is too small for real-world purposes. Let's use a data frame with 11 rows and say we want to use 5 for the analysis set, 3 for the assessment set, and leave a gap of 2 in between those two sets. We can make two such resamples from our data frame.
+
+![](images/calibration-split-window.jpg)
```
EmilHvitfeldt, line 112: could we do 4 for assessment set and a gap of 1?

right now there is very little air inside the assessment set with regards to the text

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+```
+
+
+### Index-based splitting
+
+When working with irregular series, it can be helpful to define resamples based on sliding over an index, rather than the rows directly. That index often is a form of time but it does not have to be. Let's take our previous data frame and use `x` as the index, with a small modification. The index doesn't go straight from 1 to 11 but rather is missing an observation at the index value of 2. We use the same  `lookback`, `assess_start`, and `assess_stop`. The difference is that now these define the resamples on the _index_ rather than the _rows_ directly.
+
+![](images/calibration-split-index.jpg)
+
+We still get two resamples, however, the analysis set contains only 4 rows because only those fall into the window defined by the index.
```
EmilHvitfeldt, line 250: Would it be beneficial to add the missing value at 1 or 5 such that the analysis sets have different length?

EmilHvitfeldt: or i guess, that isn't interesting at all. because then we do the same as the previous section

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+Based on `lookback`, `assess_start`, and `assess_stop`, we calculate the same allocation for the (inner) analysis set and the calibration set as in the previous example, when working on the rows directly with `sliding_window()`: three for the inner analysis set and two for the calibration set. However, the dataset we want to split only covers the range [3, 6].
+
+```{r}
+#| label: sliding_index_rsplit_2
+
+r_split <- get_rsplit(r_set, 2)
+analysis(r_split)
+```
+
+The sliding splits slide over _the data_, meaning they slide over observed values of the index and they slide only within the boundaries of the observed index values. So here, we can only slide within [3, 6] and thus cannot fit an inner analysis set of three and a calibration set of two into it. As established earlier, we fall back onto an empty calibration set in such a situation.
```
EmilHvitfeldt, line 295: we mention a couple of times that we fall back. should we mention that we fall back with a warning?

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+# absolute sizes in the inner split
+length_analysis_inner <- ceiling(prop_analysis * length_analysis_outer)
+length_calibration <- ceiling(prop_assess * length_analysis_outer)
+
+c(length_analysis_inner, length_calibration)
+```
+
+Calculating the length of the calibration set rather than the gap, together with rounding up when translating proportions to new lengths within the outer analysis set means that we prioritize allocating observations to the (inner) analysis and calibration set over allocating them to the gap. In this example here, this means that we are not leaving a gap between the analysis set and the calibration set.
+
+However, rounding up for both (inner) analysis and calibration set when we don't have a gap could mean we end up allocating more observations than we actually have. So in that case, we try to take from the calibration set if possible and thus prioritzing fitting the prediction model over the calibration model.
```
EmilHvitfeldt, line 159:
```suggestion
However, rounding up for both (inner) analysis and calibration set when we don't have a gap could mean we end up allocating more observations than we actually have. So in that case, we try to take from the calibration set if possible and thus prioritizing fitting the prediction model over the calibration model.
```

learn/work/calibration-splits/index.qmd
EmilHvitfeldt: i'm realizing that this would be a huge undertaking

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+We expect this to not be relevant in practice too often, as chances of this happening are lower with more observations and we assume you are only attempting calibration with sufficiently large datasets.
+
+
+### Period-based splitting
+
+For the third sliding function in the family, we first aggregate the index to periods and then slide across those, e.g., aggregate daily data to monthly periods. 
+
+![](images/calibration-split-period.jpg)
+
+The principle of how to contruct a calibration split on the (outer) analysis set remains the same. The challenges of abstracting away from the rows, as illustrated for sliding over observed instances of an index also remain. Here, we slide over observed periods. We observe a period, if we observe an index within that period.
```
EmilHvitfeldt, line 374:
```suggestion
The principle of how to construct a calibration split on the (outer) analysis set remains the same. The challenges of abstracting away from the rows, as illustrated for sliding over observed instances of an index also remain. Here, we slide over observed periods. We observe a period, if we observe an index within that period. 
```

EmilHvitfeldt (approved):
I think this is very valuable and high quality work.

I think this would be a good place for this content. It would also fit in rsample pkgdown for that matter.

there are small comments and nit picks but the overall structure and style i find very nice.

I'm also fine with the length. It is a complicated topic and without the prose and diagrams it would be hard to understand

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
+---
+title: "Automatic calibration splits for resamples"
+categories:
+  - post-processing
+  - calibration
+  - rsample
+type: learn-subsection
+weight: 6
+description: | 
+  Learn how tidymodels generates automatic calibration sets for post-processing.
```
topepo, line 10:
```suggestion
  Learn how tidymodels generates automatic calibration sets for postprocessing.
```

A few years ago we decided to not hyphenate this (or preprocessing)

learn/work/calibration-splits/index.qmd
topepo: I agree. I would add text to this paragraph after the first sentence and start the next paragraph with "An additional..."

learn/work/calibration-splits/index.qmd
topepo, line 40:
```suggestion
[rsample](https://rsample.tidymodels.org/) provides a collection of functions to make resamples for empirical validation of prediction models. So far, the assumption was that the prediction model is the only model that needs fitting, i.e., a resample consists of an analysis set and an assessment set.
```

Most of our other pages link to pkgdown sites when referencing the package name.

learn/work/calibration-splits/index.qmd
topepo: On Chrome, I get: 

image

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+## General principles
+
+First principle: we leave the assessment set purely for performance assessment and rather treat the analysis set as "the data to fit" - except now we split it to fit two different models instead of just one. Or in other words, we take the data for the calibration set from the analysis set, not from the assessment set. 
+
+If you compare a model with calibration to one without, and you use the same resamples, you are also using the same assessment sets.
+
+"Taking data from the analysis set" means splitting up the analysis set to end up with ... an analysis set and a calibration set. Now we have two sets called analysis set, that's confusing. If we need to distinguish them, we'll refer to them as "outer" and "inner" analysis set for "before" and "after" the split for a calibration set.
```
topepo, line 52: I found the notion of "two sets called analysis set" confusing. I think it should say that we "further split our initial analysis set into two partitions..." or something similar. I still like the inner/outer notation, but I think the process could be better worded here.

learn/work/calibration-splits/index.qmd
topepo, line 48:
```suggestion
**First principle**: we leave the assessment set purely for performance assessment and rather treat the analysis set as "the data to fit" - except now we split it to fit two different models instead of just one. Or in other words, we take the data for the calibration set from the analysis set, not from the assessment set. 
```

topepo: but I can never remember which markdown syntax we are using so maybe just one asterisk will produce bold ¯\\_(ツ)_/¯

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+#| label: calibration
+#| echo: false
+#| out.height: 350
+#| fig.align: "center"
+knitr::include_graphics("images/analysis-calibration-assessment.jpg")
+```
+
+It is important to note that including calibration requires more data than fitting a workflow without calibration. For small datasets, you might run into situations where fitting either model (primary model or calibration model) is not particularly stable - or you might not even be able to construct a calibration set properly.
+
+Second principle: we try to mimick the data splitting process used to create the (outer) analysis set and the assessment set.
```
topepo, line 65: Bold for "Second principle"

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+And lastly, the third principle: if we cannot make a calibration split, we fall back onto an empty calibration set rather than erroring.
```
topepo, line 67: same

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+Let's look at v-fold cross-validation as an example.
+
+
+### Example: v-fold cross-validation
```
topepo, line 72: For consistency, outside of syntax, we use upper case V when naming or describing this scheme.

learn/work/calibration-splits/index.qmd
topepo: LaTeX pleas

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+#| fig.align: "center"
+knitr::include_graphics("images/analysis-calibration-assessment.jpg")
+```
+
+It is important to note that including calibration requires more data than fitting a workflow without calibration. For small datasets, you might run into situations where fitting either model (primary model or calibration model) is not particularly stable - or you might not even be able to construct a calibration set properly.
+
+Second principle: we try to mimick the data splitting process used to create the (outer) analysis set and the assessment set.
+
+And lastly, the third principle: if we cannot make a calibration split, we fall back onto an empty calibration set rather than erroring.
+
```
topepo, line 68: I think that we should have a callout before the last sentence of this section that warns the users that they will need a sizable amount of data to do calibration effectively.

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+Let's take that first resample and split it into an (inner) analysis set and a calibration set. 
+
+```{r}
+#| label: sliding_window_rsplit
+
+r_split <- get_rsplit(r_set, 1)
+analysis(r_split)
+```
+
+To mimick the split into (outer) analysis and assessment set, we'll calculate the proportion of the analysis set and apply it to the observations available for fitting both the primary model and the calibration model, creating our (inner) analysis set. Since the remaining observations could be for the calibration model or a gap, we also calculate the proportion of the assessment set and use it to construct the calibration set.
```
topepo, line 132: > return the (outer) analysis set as the (inner) analsysis set

Typo + I think this should be reworded. So, ultimately, there is no calibration set, and our original analysis set remains intact? If so maybe say that? The "return" wording is confusing.

learn/work/calibration-splits/index.qmd
topepo: >  We observe a period, if we observe an index within that period.

I'd reword here too. Maybe it is comma placement but it doesn't make immediate sense to me.

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+![](images/calibration-split-period.jpg)
+
+The principle of how to contruct a calibration split on the (outer) analysis set remains the same. The challenges of abstracting away from the rows, as illustrated for sliding over observed instances of an index also remain. Here, we slide over observed periods. We observe a period, if we observe an index within that period. 
+
+For `lookback = 0`, we remain unable to make a calibration split. However, it is possible to choose a smaller period when defining the resamples, e.g., using weeks instead of months when aggregating daily data. In that case, the calibration split is also defined in terms of weeks, rather than months, and thus failing to split a single month.
+
+Now that we've looked at the principles at play here on very small datasets, let's look at a dataset with a sufficient amount of observations to consider using calibration. The `Chicago` dataset has daily ridership numbers for over 15 years. Here we are constructing resamples which use 2 weeks worth of data for the assessment set, and roughly 15 years (52*15 weeks) for the analysis set. That's enough to use some of those observations for a calibration set!
```
topepo, line 378: 💯

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+#| label: bootstrap-resample
+#| echo: false
+#| out.width: 700
+#| fig.align: "center"
+knitr::include_graphics("images/bootstrap-resample.jpg")
+```
+
+We don't want those duplicates to be split up into the inner analysis and the calibration set. Simply creating another bootstrap resample from the (outer) analysis set makes it possible for this to happen. 
+
+In the example illustration below, the third row of the (outer) analysis set is labeled `2-1`, as it is the first intance of row 2 from the original data. The other two instances of this row in the (outer) analysis set are labeled `2-2` and `2-3`. In the example, the `2-1` row gets sampled into the (inner) analysis set twice, while `2-2` and `2-3` are not sampled and thus end up in the calibration set. This means that row 2 from the original data is now in both the (inner) analysis set and the calibration set.
```
topepo, line 436:
```suggestion
In the example illustration below, the third row of the (outer) analysis set is labeled `2-1`, as it is the first intance of row 2 from the original data. The other two instances of this row in the (outer) analysis set are labeled `2-2` and `2-3`. In the example, the `2-1` row gets sampled into the (inner) analysis set twice, while `2-2` and `2-3` are not sampled and thus end up in the calibration set. This means that row 2 from the original data is now in both the (inner) analysis set and the calibration set. This is inappropriate. 
```

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+```{r}
+#| label: bootstrap-regular
+#| echo: false
+#| out.width: 700
+#| fig.align: "center"
+knitr::include_graphics("images/bootstrap-regular.jpg")
+```
+
+To avoid this, we could do a grouped resampling with the row ID as the group. Then all replications of a row would end up in the same set, either (inner) analysis or calibration. However, this would mean that rows in the calibration set are potentially not unique, unlike the typical bootstrap OOB sample.
```
topepo, line 446:
```suggestion
To avoid this, we could do a grouped resampling with the row ID as the group. Then all replications of a row would end up in the same set, either (inner) analysis or calibration. However, this would mean that rows in the calibration set are potentially not unique, unlike the typical bootstrap out-of-bag (OOB) sample.
```

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+#| label: bootstrap-modified
+#| echo: false
+#| out.width: 700
+#| fig.align: "center"
+knitr::include_graphics("images/bootstrap-modified.jpg")
+```
+
+## Summary
+
+When using calibration, we typically need generous amounts of data because we need separate data for each of these tasks: fit the primary model, fit the calibration model, and assess performance.
```
topepo, line 468:
```suggestion
When using calibration, we typically require generous amounts of data because three partitions are necessary: one for fitting the primary model, another for fitting the calibration model, and separate data to assess performance. 
```

learn/work/calibration-splits/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+When tuning or resampling a workflow with calibration, tidymodels will automatically turn classic resamples (consisting of an analysis and an assessment set) into resamples which include a calibration set in addition to the analysis and assessment set. This is done according to the following principles: 
+
+- The assessment set remains untouched, while the analysis set is split into an (inner) analysis set and a calibration set. 
+- The inner calibration split mimicks the outer split (into analysis and assesment set) as closely as possible.
+- If we can't make a calibration split based on these basic principles, we skip the calibration.
+
+For sliding splits of ordered data, applying those principles is a bit more complex than for other types of splits as the outer split into analysis and assessment is already a bit more complex. We've laid out the details of this here for reference.
+For bootstrap splits, we don't directly split the (outer) analysis set but rather sample the (inner) analysis set from the unique rows in the (outer) analysis set to avoid data leakage between (inner) analysis and calibration set.
```
topepo, line 476: Were these to be two different sentences (based on the line break)?

topepo (changes requested):

learn/work/calibration-splits/index.qmd
hfrick: Nice, thank you!

learn/work/calibration-splits/index.qmd
hfrick: Okay great, I've updated it elsewhere in the doc as well

PR #148, opened 2026-03-10
split up make_function_list.R

EmilHvitfeldt:
This PR refactors `make_function_lists.R` script into one script for each table corresponding to the tables seen on https://www.tidymodels.org/find/.

- Split `make_function_lists.R` into separate scripts in `make_function_lists/`:
  - `_utils.R` - shared utilities and caching logic
  - `broom.R`, `recipes.R`, `tidymodels.R`, `parsnip.R`, `sparse.R`, `tidyclust.R` - one for each table
  - `run_all.R` - runs all of them (same as running make_function_list.R)

I replace `pkgdown::as_pkgdown()` with direct Rd file parsing via `tools::parse_Rd()`.

The more important change is the implementation of caching. We now store a cache of the things we are fetching in `.pkg_cache.csv`. Reducing the total time taken to run all of this down to around 3 minutes.
Notice we are still fetching things from websites, so this is still not static.


EmilHvitfeldt:
I don't know if this is too much or not. It increases the complexity

PR #244, opened 2026-05-13
Efficient grid search article

topepo:
To replace the unfinished one in #86 

Also added a config file for panache. 

@EmilHvitfeldt Do I just commit the qmd here?


learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+tidymodels has three possible components to add to a model pipeline (a.k.a. a *workflow*):
+
+- Preprocessors: computations that prepare data for the model
+- Supervised model: the actual model fit
+- Postprocessing: adjustments to model predictions
+
+Each of these stages can have tuning parameters. Some examples:
+
+- We might add a spline basis expansion to a predictor to enable to have a nonlinear trend in the model. We take one predictor and create multiple columns based on it. As we add new spline columns, the model can be more complex. We don't know how many columns to add, so we try different values and see which maximizes performance.
```
hfrick, line 96: one of "enable" or "have" might be enough in this sentence

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+  ggplot(aes(neighbors, min_n)) + 
+  geom_point() + 
+  labs(x = "# Neighbors", y = "Min N")
+```
+
+To tune our model, we would iterate over this grid, fitting six different models and evaluating their performance to choose the best one.
+
+There's a problem, though: UMAP is a very expensive preprocessing method. If we loop through the rows of our grid, we'll end up estimating UMAP times even though there are only two possible values. That's incredibly inefficient.
+
+### Problem: DRY (Don't Repeat Yourself)
```
hfrick, line 153: The negation here is potentially confusing/misleading. Repetition is the problem, not the "don't"/avoidance thereof.

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+If we have an inexpensive preprocessor for our data, repeating it is not a big problem; we would probably just keep things simple and effectively use a `for` loop over our rows.
+
+However, there is a much better solution to make the computations as inexpensive as possible.
+
+### Solution: Stagewise Estimation and Conditional Execution
+
+We can separate our hyperparameters into three possible sets for the three possible stages of the pipeline: preprocessing, the supervised model, and postprocessing. In our example, we have a single tuning parameter for preprocessing and the supervised model.
```
hfrick:
```suggestion
We can separate our hyperparameters into three possible sets for the three possible stages of the pipeline: preprocessing, the supervised model, and postprocessing. In our example, we have a single tuning parameter each for preprocessing and the supervised model.
```

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+iter_1 + iter_2 + iter_3
+```
+
+We end up training six different tree ensembles but estimate only two UMAP models. This strategy can save a lot of time, especially for regular grids.
+
+This conditional execution strategy seems like a simple solution, and it is. However, it will become more complex if we recursively iterate over the model and post-processing parameters.
+
+## Problem: Curse of Dimensionality
+
+Part of the problem with the conventional wisdom that "grid search is massively inefficient" is the Curse of Dimensionality. As the problem dimensions increase, the volume of the search space becomes vast. In our case, this means that, as we add more tuning parameters, the parameter space explodes exponentially. For example, if we wanted to evaluate four values of each of $p$ tuning parameters, we would need $p^d$ combinations. With four parameters that are evaluated on five distinct points, the grid would have 1,024 rows.
```
hfrick:
```suggestion
Part of the problem with the conventional wisdom that "grid search is massively inefficient" is the Curse of Dimensionality. As the problem dimensions increase, the volume of the search space becomes vast. In our case, this means that, as we add more tuning parameters, the parameter space explodes exponentially. For example, if we wanted to evaluate $d$ values of each of $p$ tuning parameters, we would need $d^p$ combinations. With five parameters that are evaluated on four distinct points, the grid would have 1,024 rows.
```

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+It's a difficult situation as long as you don't think about the problem very hard. There are two ways to avoid this issue:
+
+- [Parallelize](http://tmwr.org/grid-search#parallel-processing) the tasks. If I can do 10 things at once, the cost of doing more drops significantly. Grid search is an "embarrassingly parallel" problem, and running the computations in parallel can produce speedups of several-fold. This is not hard to do in R or tidymodels.
```
hfrick:
```suggestion
- [Parallelize](http://tmwr.org/grid-search#parallel-processing) the tasks. If we can do 10 things at once, the cost of doing more drops significantly. Grid search is an "embarrassingly parallel" problem, and running the computations in parallel can produce speedups of several-fold magnitude. This is not hard to do in R or tidymodels.
```

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+  )  +
+  labs(y = NULL, title = "1 Neighbor, Min N 100")
+iter_1 + iter_2 + iter_3
+```
+
+We end up training six different tree ensembles but estimate only two UMAP models. This strategy can save a lot of time, especially for regular grids.
+
+This conditional execution strategy seems like a simple solution, and it is. However, it will become more complex if we recursively iterate over the model and post-processing parameters.
+
+## Problem: Curse of Dimensionality
```
hfrick, line 271: Should this be a subheading under Scenario A?

hfrick: (Same for the corresponding Solution heading)

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+We could go back to a regular grid. That will be very efficient for the submodel parameters (assuming that there is one), but very inefficient for the other parameters.
+
+Another option is a mixture of space-filling points and a regular grid.
+
+### Solution: Mixed Grids
+
+A mixed grid would create a space-filling design for the parameters that cannot exploit submodels and then, for each design point in that grid, create a dense sequence of values for the submodel parameter. This may significantly increase the number of grid points evaluated, but limits the number of models that are fit.
+
+Here's an example with many more grid points (900), but trains on the same number of grid points as the 3D grid in the previous section (30).
```
hfrick:
```suggestion
Here's an example with many more grid points (900), but trains on the same number of models as the 3D grid in the previous section (30).
```

hfrick: or workflows instead of models

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+## Postprocessing
+
+Postprocessing methods transform predictions to make them more useful. We can think of them as being estimated or not. As was mentioed in the first section, if we have a classification model with two possible classes, we might want to adjust the rule that converts the predicted class probabilities to "event" or "no event". This enables us to optimize for what we care about (e.g., sensitivity or specificity). For this example, we are not estimating a parameter to test a potential threshold; we simply apply something like `ifelse(prob >= threshold, "event", "no event")` to transform the class predictions.
```
hfrick:
```suggestion
Postprocessing methods transform predictions to make them more useful. We can think of them as being either estimated or not. As was mentioned in the first section, if we have a classification model with two possible classes, we might want to adjust the rule that converts the predicted class probabilities to "event" or "no event". This enables us to optimize for what we care about (e.g., sensitivity or specificity). For this example, we are not estimating a parameter to test a potential threshold; we simply apply something like `ifelse(prob >= threshold, "event", "no event")` to transform the class predictions.
```

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+An example of an estimated postprocessor is calibration. Here, we use a model to estimate when we are underpredicting the outcome, and then make a correction to eliminate those errors. For example, in a regression model, if we are consistently underpredicting samples whose outcomes are at the high end of the distribution, we can estimate that error and subtract it when we make predictions in that range. It's actually more complicated than that, but this explanation gives you what you need to know for this discussion.
```
hfrick:
```suggestion
An example of an estimated postprocessor is calibration. Here, we use a model to estimate when we are underpredicting the outcome, and then make a correction to eliminate those errors. For example, in a regression model, if we are consistently underpredicting samples whose outcomes are at the high end of the distribution, we can estimate that error and add it when we make predictions in that range. It's actually more complicated than that, but this covers the ideas relevant for this discussion.
```

learn/work/efficient-grid-search/index.qmd
```diff
@@ -0,0 +1,478 @@
...
+
+The most efficient way to execute these computations when we need to tune the postprocess is another case of conditional execution. For each combination of modeling and preprocessing parameters, we loop over the postprocessing tuning parameters to estimate (if needed) and apply the transformations.
+
+The only real speed-up comes when the postprocessor is *not* estimated. In that case, there is a slight efficiency boost by storing all predictions in a data frame and applying the transformation to them all at once. For example, if I tuned by probability threshold over 50 values, I can row-bind them into a single data frame and use an efficient in-line application of `ifelse()`. It just saves us a `for` loop, but these small efficiencies can add up.
```
hfrick:
```suggestion
The only real speed-up comes when the postprocessor is *not* estimated. In that case, there is a slight efficiency boost by storing all predictions in a data frame and applying the transformation to them all at once. For example, if we tuned a probability threshold over 50 values, we can row-bind them into a single data frame and use an efficient in-line application of `ifelse()`. It just saves us a `for` loop, but these small efficiencies can add up.
```

learn/work/efficient-grid-search/index.qmd
hfrick: everything else is "we"

hfrick (approved):
This is really neat and clearly laid out. Do we make use of mixed grids inside of `tune_grid()`? If not, then I would mention that this is something for the user to do outside of `tune_grid()`, based on how expensive evaluation of their model of choice is. 

Other than that I mostly added copy-edits as suggestions.

hfrick:
The [README](https://github.com/tidymodels/tidymodels.org#contributing) says to include the rendered files. You could do it locally or try the new `/render` command here in the PR 👀

Issue #257, opened 2026-05-21
new Learn article: "X tools for visualizing decision trees"


topepo:
There is `plot.rpart()`, `plot.as.party()` (the best one), the rpart.plot package, [parttree](https://grantmcdermott.com/parttree/index.html)::[geom_parttree.R](https://github.com/grantmcdermott/parttree/blob/main/R/geom_parttree.R), and probably others. 

A little gallery and discussion would be nice.

PR #270, opened 2026-05-26
Add blog post on tree visualization

CharlotteZZh:
working on #257


learn/models/tree-visualization/index.html.md
```diff
@@ -0,0 +1,60 @@
...
+type: learn-subsection
+weight: 6
+description: | 
+  Visualize trees.
+  
+toc: true
+toc-depth: 2
+r-packages:
+  - tidymodels
+  - pls
```
EmilHvitfeldt: DEMO:

oops we didn't need pls

learn/models/tree-visualization/index.qmd
```diff
@@ -0,0 +1,295 @@
...
+pkgs <- unlist(rmarkdown::metadata[["r-packages"]])
+
+theme_set(theme_bw() + theme(legend.position = "top"))
+```
+
+
+## Introduction
+`r article_req_pkgs(pkgs)`
+
+Decision trees are widely used because they are easy to interpret and can
```
topepo: I'd avoid wrapping paragraphs.

learn/models/tree-visualization/index.qmd
```diff
@@ -0,0 +1,295 @@
...
+capture nonlinear relationships between predictors and outcomes. Once a
+tree model has been fit, visualization becomes an important step for
+understanding splits, predictions, and decision boundaries.
+
+R has several different tools for plotting trees. Some produce simple
+base R graphics, while others focus on cleaner layouts or integration
+with `ggplot2`. This article walks through a few common approaches and
```
topepo: We follow the tidyverse style guide, which states that package names are not formatted. I don't particularly like this, but removing the backticks will keep this post consistent with the others.

learn/models/tree-visualization/index.qmd
```diff
@@ -0,0 +1,295 @@
...
+Let's compares and discusses several common methods for visualizing decision
+trees in R:
+
+  - [`plot.rpart()`](#base-r-visualization-with-plotrpart)
+  - [`partykit::plot()`](#visualization-with-partykit)
+  - [`rpart.plot::rpart.plot()`](#visualization-with-rpartplot)
+  - [`parttree::geom_parttree()`](#visualization-with-parttree)
+  - [`rattle::fancyRpartPlot()`](#visualization-with-fancyrpartplot)
+  
+We will fit a simple classification tree using the `iris` data set and
```
topepo: Can we use the penguins data instead? We try to avoid Iris because of its roots in eugenics journals.

learn/models/tree-visualization/index.qmd
```diff
@@ -0,0 +1,295 @@
...
+  ip_args = list(abbreviate = FALSE),
+  tp_args = list(id = FALSE)
+)
+```
+
+In this example:
+
+- `abbreviate = FALSE` prevents split labels from being shortened
+- `id = FALSE` removes terminal node numbers for a cleaner display
+
```
topepo, line 153: Another nice option to mention is `ep_args = list(justmin = 15)`, which vertically offsets the rules. So if you have long rules like `town %in% vector_of_ten_towns`, the labels on top of the lines won't overlap.

topepo (commented):
Looks good! The writing is very clear and easy to read.

learn/models/tree-visualization/index.qmd
EmilHvitfeldt: Specifically written here: https://style.tidyverse.org/documentation.html#package-names

learn/models/tree-visualization/index.qmd
EmilHvitfeldt: and we have that data set to be loaded from `library(modeldata)`

EmilHvitfeldt:
another thing that might be worth looking into are regression style decision trees. not necessarily to double the length of the post. but there might be some differences worth highlighting.

also would not be surprised if one of these packages only supported classification trees 🙃

CharlotteZZh:
Updated a few things based on your feedback

EmilHvitfeldt:
Thank you! this is coming along really nicely, one last thing i have is that the header titles for the example sections looks a little off.

I think we can remove the "Visualization with " from all of them and it will still flow nicely

Screenshot 2026-05-28 at 10 13 57 AM

CharlotteZZh:
> Thank you! this is coming along really nicely, one last thing i have is that the header titles for the example sections looks a little off.
> 
> I think we can remove the "Visualization with " from all of them and it will still flow nicely
> 
> Screenshot 2026-05-28 at 10 13 57 AM

updated! thanks

Issue #254, opened 2026-05-21
new Learn article: model monitoring using slider


topepo:
Once a model is built and new data comes in, users want to see if performance drifts over time. 

We should have an article that largely uses the slider and purr packages to take timestamped data and 

- Predicts the data
- splits the data over time intervals (i.e. check performance every two weeks)
- computed metrics
- plots them over time
etc.

PR #274, opened 2026-05-29
article on model monitoring using slider

CharlotteZZh:
#254


learn/models/model-monitoring-using-slider/index.qmd
```diff
@@ -0,0 +1,415 @@
...
+    outcome ~ x1 + x2,
+    data = train_data
+  )
+
+log_fit
+```
+
+## Generating predictions
+
+As new observations arrive, the deployed model generates predictions. Each observation now contains: a timestamp, the true outcome, the predicted class, predicted probabilities. This structure is common in production monitoring systems and forms the basis of our analysis.
```
EmilHvitfeldt:
```suggestion
As new observations arrive, the deployed model generates predictions. Each observation now contains: a timestamp, the true outcome, the predicted class, and predicted probabilities. This structure is common in production monitoring systems and forms the basis of our analysis.
```

learn/models/model-monitoring-using-slider/index.qmd
```diff
@@ -0,0 +1,415 @@
...
+  - element 2 contains observations from the next window, and so on.
+  
+This list will serve as the foundation for our monitoring workflow.
+
+## Understanding `slide_index()`
+
+The `slide_index()` function has several important arguments.
+Here:
+
+  - .x specifies the object being sliced,
```
EmilHvitfeldt: Since these are arguments we should put them as inline code.

```suggestion
  - `.x` specifies the object being sliced,
```

same should be applied everywhere else

learn/models/model-monitoring-using-slider/index.qmd
EmilHvitfeldt: > Each observation now contains: a timestamp, the true outcome, the predicted class, predicted probabilities.

It technically also contains the predictors as well. this is properly fine as we didn't write "only contains" but might be worth adding.

Also I think it is worthwhile to mention that we used `augment()` over `predict()` as we get all these columns together in a safe way

learn/models/model-monitoring-using-slider/index.qmd
EmilHvitfeldt: like with `.before = 13` below which should be inside `

learn/models/model-monitoring-using-slider/index.qmd
```diff
@@ -0,0 +1,415 @@
...
+  - .i specifies the index variable,
+  - .before controls the size of the rolling window: We set .before = 13 because slider includes the current observation in each window. As a result, each window contains the current day plus the previous 13 days, for a total of 14 days.
+  - .complete = TRUE removes incomplete windows,
+  - .f specifies the function applied to each window.
+  
+Because we use `identity()`, slider returns each window unchanged. This creates a collection of evaluation periods that we can analyze independently.
+
+## Computing metrics with purrr
+
+In practice, monitoring systems rarely track a single metric. Suppose we want both accuracy and ROC AUC.
```
EmilHvitfeldt: text mentions accuracy and ROC AUC but the show also include mn_log_loss

learn/models/model-monitoring-using-slider/index.qmd
```diff
@@ -0,0 +1,415 @@
...
+      accuracy_vec(
+        dat$outcome,
+        dat$.pred_class
+      ),
+
+    roc_auc =
+      roc_auc_vec(
+        dat$outcome,
+        dat$.pred_yes,
+        event_level = "second"
```
EmilHvitfeldt: instead of setting `event_level = "second"`, we should just update the levels at the top of the data

learn/models/model-monitoring-using-slider/index.qmd
```diff
@@ -0,0 +1,415 @@
...
+
+## Why use time windows
+
+Looking at predictions one at a time is usually not very informative. If a model makes a mistake today, that does not necessarily mean anything is wrong. Instead, we generally summarize performance across a larger set of observations. For example:
+
+Window | Dates
+-------|------------------
+1      | Jan 1 - Jan 14
+2      | Jan 15 - Jan 28
+3      | Jan 29 - Feb 11
```
EmilHvitfeldt: This table doesn't render for me

Image

learn/models/model-monitoring-using-slider/index.qmd
```diff
@@ -0,0 +1,415 @@
...
+
+monitoring_results <-
+  windows %>%
+  map(compute_metrics) %>%
+  list_rbind()
+monitoring_results %>%
+  slice_head(n = 5)
+```
+
+## Visualizing accuracy through time
```
EmilHvitfeldt: These three sections feels a little too long for me. I think we should combine them as a faceted ggplot. Since they all show the same thing just in different ways

EmilHvitfeldt: Also while I do see an decrease in performance is it not THAT noticeable since the variance of the performance appears quite high IMO

EmilHvitfeldt (commented):
Good first draft!

I added some suggestions and such. I also noticed a couple of unnamed-chunk files. we should properly delete them just in case

learn/models/model-monitoring-using-slider/index.html.md
```diff
@@ -0,0 +1,501 @@
...
+```
+:::
+
+::: {.cell layout-align="center"}
+
+```{.r .cell-code}
+# We fit a logistic regression model:
+log_fit <-
+  logistic_reg() %>%
+  set_engine("glm") %>%
```
topepo: This is the default engine so you can exclude this line

learn/models/model-monitoring-using-slider/index.html.md
```diff
@@ -0,0 +1,501 @@
...
+Metrics computed within these windows are typically more stable and easier to interpret than metrics based on individual predictions. The next challenge is creating these windows efficiently and applying the same calculations to each one. This is where the slider package becomes useful.
+
+## Creating rolling windows with slider
+
+The basic idea behind slider is straightforward: apply the same function repeatedly across a series of moving windows. In our monitoring workflow, each window represents a rolling 14-day period of observations. We start by creating these windows using the observation date as the time index.
+
+::: {.cell layout-align="center"}
+
+```{.r .cell-code}
+windows <- slide_index(
```
topepo: I would use `sliding_period()` so that it can handle non-sequential time values, and order won't matter. Something like this (using a built-in data set from modeldata):

```r
Chicago |>
  sliding_period(
    index = "date",  
    period = "week",
    every = 2, 
    skip = 1
  ) |> 
  mutate(
    date_start = map_vec(splits, ~ min(analysis(.x)$date)), 
    date_stop  = map_vec(splits, ~ max(analysis(.x)$date)),
    num_values = map_int(splits, ~ nrow(analysis(.x)))
  )
```

This will give you a data frame with a `splits` column that you can use to pull out the data as shown ☝️

learn/models/model-monitoring-using-slider/index.html.md
```diff
@@ -0,0 +1,501 @@
...
+Because we use `identity()`, slider returns each window unchanged. This creates a collection of evaluation periods that we can analyze independently.
+
+## Computing metrics with purrr
+
+In practice, monitoring systems rarely track a single metric. Suppose we want both accuracy and ROC AUC.
+
+::: {.cell layout-align="center"}
+
+```{.r .cell-code}
+compute_metrics <- function(dat) {
```
topepo: A _much_ easier way to do this is to use a metric set, like 

```r
metric_fn <- metric_set(accuracy, roc_auc, mn_log_loss_vec)
```

and will stack the metrics vertically.

EmilHvitfeldt:
looking great!

